PR #12, opened 2017-10-03
Feature/edit profile widget

olslash:
@atungw 

this PR adds AppBar and a few other components needed for the edit-profile widget, plus some associated config changes as i add features to the veritone-react-common package

(still WIP)


olslash:
merging so work can be based off this foundation

PR #36, opened 2018-01-26
Add ocr pill with query for exact text search

AlexOVeritone:
1. Add Recognized Text pill component that does the exact text search.
2. Correct css to place pills in row - currently stacks in column
3. Change transcript modal text to be the same as in discovery-app

### DIFF
--- a/packages/veritone-react-common/src/components/SearchBar/story.js
+++ b/packages/veritone-react-common/src/components/SearchBar/story.js
@@ -8,6 +8,12 @@ import {
   TranscriptDisplay,
   TranscriptConditionGenerator
 } from 'components/TranscriptSearchModal';
+import {
+  RecognizedTextSearchModal,
+  RecognizedTextDisplay,
+  RecognizedTextConditionGenerator
+} from 'components/RecognizedTextSearchModal';
+
 import SearchBarContainer from './SearchBarContainer';
 import { SearchBar } from '.';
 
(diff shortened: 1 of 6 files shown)


packages/veritone-react-common/src/components/SearchBar/story.js
```diff
@@ -20,15 +26,29 @@ const transcript = {
   showPill: true
 };
 
+const recognizedText = {
```
SteveShaffer: @atungw Why are we doing another key for each engine category?  Why not just add these attributes to `engineCategoryMapping` below and pass the engine category IDs as the `enabledEngineCategories`?

packages/veritone-react-common/src/components/TranscriptSearchModal/index.js
```diff
@@ -67,7 +67,7 @@ export default class TranscriptSearchModal extends React.Component {
 export const TranscriptSearchForm = ( { defaultValue, cancel, onSubmit, onChange, onKeyPress, inputValue } ) => {
   return (
   <div>
-    <DialogTitle>Search Transcript</DialogTitle>
+    <DialogTitle>Search by Keyword</DialogTitle>
```
SteveShaffer: I'm not convinced we shouldn't take this opportunity to "generalize" this stuff.

Let's leave this for now.  But at the end of this I'll probably take a pass through here and tweak the copy to be more in line with our generalized, non-media-specific data model.

atungw: FYI, just recapping our last conversation but most of that information is defined here and will be passed into the SearchBar to configure it.

See the endpoint: https://enterprise.stage.veritone.com/api/engine/category?time=1517268957867

packages/veritone-react-common/src/components/SearchBar/story.js
atungw: This stuff is hardcoded for convenience, but it'll mostly be configured from here. https://enterprise.stage.veritone.com/api/engine/category?time=1517268957867

PR #55, opened 2018-03-13
Add basic engine output selector tabs.

AlexOVeritone:


packages/veritone-widgets/src/redux/modules/mediaDetails/saga.js
```diff
@@ -149,7 +152,7 @@ function* loadEngineCategoriesSaga(widgetId, mediaId, callback = noop) {
         return;
       }
       if (category.engines.some(engine => engine.status !== 'complete')) {
-        category.status = 'in_progress';
+        category.status = 'inprogres';
```
ChrisPelletier: is this a typo?

packages/veritone-react-common/src/components/MediaDetails/index.js
```diff
@@ -12,13 +13,23 @@ export default class MediaDetails extends Component {
   };
 
   state = {
-    selectedEngineCategoryId: ''
+    selectedEngineCategory: this.props.engineCategories && this.props.engineCategories.length ? this.props.engineCategories[0] : null
```
ChrisPelletier: Is this something we can move to redux?

packages/veritone-widgets/src/redux/modules/mediaDetails/saga.js
AlexOVeritone: I think I saw tried status without _ in some documentation, however, couldn't find that documentation.

packages/veritone-react-common/src/components/MediaDetails/index.js
AlexOVeritone: yes, and it would feed it down the stack I think

PR #56, opened 2018-03-13
engine selection widget

charliemantle:
initial version of engine selection widget


packages/veritone-widgets/src/widgets/EngineSelection/EngineListView/EngineSelectionRow/index.js
```diff
@@ -0,0 +1,134 @@
...
+
+import ToggleButton from '../../ToggleButton/';
+
+import styles from './styles.scss';
+
+@connect(
+  (state, ownProps) => ({
+    isSelected: engineSelectionModule.engineIsSelected(
+      state,
+      ownProps.engineId
```
olslash: proptypes, required

packages/veritone-widgets/src/widgets/EngineSelection/EngineListView/EngineSelectionRow/index.js
```diff
@@ -0,0 +1,134 @@
...
+    isSelected: bool.isRequired,
+    isChecked: bool.isRequired,
+    onViewDetail: func.isRequired,
+    addEngines: func.isRequired,
+    removeEngines: func.isRequired,
+    checkEngine: func.isRequired,
+    uncheckEngine: func.isRequired
+  };
+
+  handleOnChange = () => {
```
olslash: this is just a stylistic issue, but these should be `handleChange`, `handleClick` (saw this issue in a few places). Here's the eslint rule to prove i'm not just making this up: https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-handler-names.md (maybe we should add it)

PR #63, opened 2018-03-19
Feature/vtn 6543 mdp info panel read mode

AlexOVeritone:


packages/veritone-react-common/src/components/MediaDetails/styles.scss
```diff
@@ -82,11 +77,11 @@
 
       .toEditModeButton {
         text-transform: uppercase;
-        background-color: #2196F3;
-        color: #FFFFFF;
+        background-color: #2196f3;
```
ChrisPelletier: something to consider but not critical. There are scss variables under /src/styles/modules/_variables.scss with colors like $blue-1

PR #68, opened 2018-03-22
Feature/vtn 6758 logo recognition engine output component

KhuongVeritone:
Add Logo Recognition Output Component

### DIFF
--- a/packages/veritone-react-common/src/components/LogoDetectionEngineOutput/index.js
+++ b/packages/veritone-react-common/src/components/LogoDetectionEngineOutput/index.js
@@ -119,8 +119,8 @@ export default class LogoDetectionEngineOutput extends Component {
      * @return MUI Select Element
      */
     drawEngineSelection (data) {
-        if (!data || data.length === 0)                         return '';
-        if (!this.state.sourceEngineId && data.length > 0)      this.state.sourceEngineId = data[0].sourceEngineId;
+        if (!data || data.length === 0)      return '';
+        if (!this.state.sourceEngineId)      this.state.sourceEngineId = data[0].sourceEngineId;
         
         return (
             <Select value={this.state.sourceEngineId} onChange={this.onEngineChanged}>
--- a/packages/veritone-react-common/src/components/LogoDetectionEngineOutput/index.js
+++ b/packages/veritone-react-common/src/components/LogoDetectionEngineOutput/index.js
@@ -8,6 +8,7 @@ import Select from 'material-ui/Select';
 import { MenuItem } from 'material-ui/Menu';
 import EngineOutputHeader from '../EngineOutputHeader';
 import PillButton from '../Parts/Buttons/PillButton';
+import { msToReadableString } from '../../helpers/time'
 import styles from './styles.scss';
 
 export default class LogoDetectionEngineOutput extends Component {
@@ -129,8 +130,8 @@ export default class LogoDetectionEngineOutput extends Component {
                     // Draw detected logos in the series
                     engineInfo.series.map((itemInfo, index) => {
                         if (itemInfo.logo) {                                                            //Look for detected logo
-                            let startTime = this.formatTime(itemInfo.startTimeMs);
-                            let endTime = this.formatTime(itemInfo.endTimeMs);
+                            let startTime = msToReadableString(itemInfo.startTimeMs);
+                            let endTime = msToReadableString(itemInfo.endTimeMs);
                             console.log(itemInfo.startTimeMs, currentPlayTime);
                             let logoItem = (
                                 <PillButton
@@ -152,22 +153,6 @@ export default class LogoDetectionEngineOutput extends Component {
             });
         }
         return items;
-    }
-
-    formatTime (millisec) {
-        let dateObject = new Date(millisec);
-        let numDays = dateObject.getUTCDate() - 1;
-        let numHours = dateObject.getUTCHours() + numDays * 24;
-        let numMinites = dateObject.getUTCMinutes();
-        let numSeconds = dateObject.getUTCSeconds();
-        let numMilliseconds = dateObject.getUTCMilliseconds();
-
-        let formatedString = '';
-        formatedString = formatedString + numHours.toLocaleString(undefined, {minimumIntegerDigits: 2}) + ':';
-        formatedString = formatedString + numMinites.toLocaleString(undefined, {minimumIntegerDigits: 2}) + ':';
-        formatedString = formatedString + numSeconds.toLocaleString(undefined, {minimumIntegerDigits: 2}) + ':';
-        formatedString = formatedString + numMilliseconds.toLocaleString(undefined, {minimumIntegerDigits: 3});
-        return formatedString;
     }
 
     //================================================================================
--- a/packages/veritone-react-common/src/components/LogoDetectionEngineOutput/index.js
+++ b/packages/veritone-react-common/src/components/LogoDetectionEngineOutput/index.js
@@ -13,7 +13,7 @@ import styles from './styles.scss';
 export default class LogoDetectionEngineOutput extends Component {
     static propTypes = {
         label: PropTypes.string,
-        data: PropTypes.array,
+        data: PropTypes.arrayOf(Object),
         className: PropTypes.string,
         headerClassName: PropTypes.string,
         bodyClassName: PropTypes.string,
@@ -36,7 +36,6 @@ export default class LogoDetectionEngineOutput extends Component {
             isExpanded: false
         };
 
-        this.onItemSelected = this.onItemSelected.bind(this);
         this.onEngineChanged = this.onEngineChanged.bind(this);
         this.onExpandViewClicked = this.onExpandViewClicked.bind(this);
     }
@@ -44,9 +43,9 @@ export default class LogoDetectionEngineOutput extends Component {
     //================================================================================
     //=============================Interaction Functions==============================
     //================================================================================
-    onItemSelected (value) {
+    onItemSelected = item => (event) => {
         let itemInfo = {
-            item: value,
+            item: item,
             sourceEngineId: this.state.sourceEngineId || this.props.data[0].sourceEngineId
         }
         
@@ -121,15 +120,16 @@ export default class LogoDetectionEngineOutput extends Component {
                             let startTime = this.formatTime(itemInfo.startTimeMs);
                             let endTime = this.formatTime(itemInfo.endTimeMs);
                             let logoItem = (
-                                <PillButton 
+                                <PillButton
+                                    value={index} 
                                     label={itemInfo.logo.label} 
                                     info={startTime + ' - ' + endTime}
                                     className={classNames(styles.item, itemClassName)}
                                     labelClassName={classNames(styles.label, itemLabelClassName)}
                                     infoClassName={itemInfoClassName}
                                     key={'logo-' + engineInfo.sourceEngineId + index}
-                                    onClick={(event) => {this.onItemSelected(itemInfo)}}
-                                ></PillButton>
+                                    onClick={this.onItemSelected(itemInfo)}
+                                />
                             );
                             items.push(logoItem);
                         }
--- a/packages/veritone-react-common/src/components/LogoDetectionEngineOutput/index.js
+++ b/packages/veritone-react-common/src/components/LogoDetectionEngineOutput/index.js
@@ -1,0 +1,193 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+
+import classNames from 'classnames';
+import styles from './styles.scss';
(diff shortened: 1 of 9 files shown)


packages/veritone-react-common/src/components/LogoDetectionEngineOutput/index.js
```diff
@@ -0,0 +1,196 @@
...
+        data: PropTypes.array,
+        headerClassName: PropTypes.string,
+        bodyClassName: PropTypes.string,
+        itemClassName: PropTypes.string,
+        itemLabelClassName: PropTypes.string,
+        itemInfoClassName: PropTypes.string,
+        onItemSelected: PropTypes.func,
+        onEngineSelected: PropTypes.func,
+        onExpandViewClicked: PropTypes.func
+    };
```
ChrisPelletier: We need to also be able to pass in something like mediaPlayerTime or mediaPlayerPosition that indicates where the video is at so we can highlight the logo pills where media player time falls between start and end time. see https://veritone.invisionapp.com/d/main#/console/13611461/286589654/preview

packages/veritone-react-common/src/components/Parts/Buttons/PillButton/styles.scss
```diff
@@ -0,0 +1,37 @@
...
+	}
+
+	.label {
+		white-space: nowrap; 
+		overflow: hidden;
+		text-overflow: ellipsis;
+	}
+	
+	.info {
+		color: #2196f3;
```
ChrisPelletier: If you @import 'src/styles/modules/variables'; there is a $blue-1 variable you can use here

KhuongVeritone:
@charliemantle yeah, I'll try to enable it again. I thought it's already on but somehow it didn't catch anything. Btw, which parts did you find are inconsistent? Is it the double quote in mock data or the tab vs 2 spaces in the js file? I'm still not sure about the format we are using.

packages/veritone-react-common/src/components/LogoDetectionEngineOutput/story.js
```diff
@@ -0,0 +1,103 @@
...
+	);
+});
+
+const mockData = [
+	{
+		startTimeMs: 0,
+		endTimeMs: 9000,
+		sourceEngineId: '2dc5166f-c0ad-4d84-8a85-515c42b5d357',
+		sourceEngineName: 'Server-R',
+		taskId: 'e1fa7d7c-6f1c-480e-b181-68940509f070-fef496da-f36e-49ec-a304-426d96017ddf',
```
SteveShaffer: I don't think we can assume this chunk of data is all associated to a single task.  In general, we don't want MDP or engine output viewers to know about either assets or tasks, just take time series data.  @AlexOVeritone any insight here?  I'll review the GraphQL branch in a bit to see what the connection is.

packages/veritone-react-common/src/components/LogoDetectionEngineOutput/index.js
```diff
@@ -0,0 +1,198 @@
...
+		mediaPlayerTime: -1
+	};
+
+	constructor(props) {
+		super(props);
+		this.state = {
+			isExpanded: false
+		};
+
+		this.onEngineChanged = this.onEngineChanged.bind(this);
```
SteveShaffer: As discussed on Friday with @ChrisPelletier moving these selectors into a higher-level component keeps the props for this component much simpler and allows us to consolidate these selectors into a engine output viewer container component.

But if all other engine output components are currently doing this and we don't want to refactor them all right now, consistency wins.

packages/veritone-react-common/src/components/LogoDetectionEngineOutput/index.js
```diff
@@ -0,0 +1,198 @@
...
+		if (this.props.onItemSelected) this.props.onItemSelected(itemInfo);
+	}
+
+	onEngineChanged(event) {
+		let selectedID = event.target.value;
+		this.setState({
+			sourceEngineId: selectedID
+		});
+
+		if (this.props.onEngineSelected) this.props.onEngineSelected(selectedID);
```
SteveShaffer: Usual Veritone code style is
```
if () {
  // always drop onto a second line
}
```
...unless I've been out of touch for too long and someone like @olslash or @ChrisPelletier corrects me.

Please correct all over.  I'm actually surprised prettier doesn't enforce something like that.

packages/veritone-react-common/src/components/LogoDetectionEngineOutput/index.js
```diff
@@ -0,0 +1,198 @@
...
+		let itemLabelClassName = this.props.itemLabelClassName;
+		let selectedEngineID = this.state.sourceEngineId || data[0].sourceEngineId;
+
+		let items = [];
+		if (data && data.length > 0) {
+			data.map((engineInfo) => {
+				if (engineInfo.sourceEngineId === selectedEngineID) {									//Look for selected engine
+
+					// Sort detected logos by there start time and end time
+					engineInfo.series.sort(function (item1, item2) {
```
SteveShaffer: To me, it's much easier to read
```
import {sortBy} from lodash;
engineInfo.series = sortBy(engineInfo.series, 'startTimeMs', 'endTimeMs');
```
People gots all sorts of opinions on lodash vs. vanilla JS so I'm fine to defer to the consensus on this one but I do see lodash in package.json so I think it's fine to use here.

packages/veritone-react-common/src/components/Parts/Buttons/PillButton/index.js
```diff
@@ -0,0 +1,60 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+import defaultStyles from './styles.scss';
+
+export default class PillButton extends Component {
+	static propTypes = {
```
SteveShaffer: Indent should be 2 spaces?  Is prettier setup right?

packages/veritone-react-common/src/components/Parts/Buttons/PillButton/index.js
```diff
@@ -0,0 +1,60 @@
+import React, { Component } from 'react';
```
SteveShaffer: Kinda surprised to see PillButton in here if we've already done face and object.  Shouldn't it be using the same thing?

packages/veritone-react-common/src/components/Parts/Buttons/story.js
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+
+import styles from './story.styles.scss';
+import PillButton from './PillButton';
+
+storiesOf('Common Buttons', module).add('Pill Button', () => {
+	return (
```
SteveShaffer: Same indentation comment as above

packages/veritone-react-common/src/components/LogoDetectionEngineOutput/index.js
```diff
@@ -0,0 +1,198 @@
...
+		};
+
+		this.onEngineChanged = this.onEngineChanged.bind(this);
+		this.onExpandViewClicked = this.onExpandViewClicked.bind(this);
+	}
+
+	//================================================================================
+	//=============================Interaction Functions==============================
+	//================================================================================
+	onItemSelected = item => (event) => {
```
olslash: this technique (currying) is a clever way around the jsx-no-bind eslint rule, but it has the same negative effect that we try to avoid with that rule -- basically this function returns a new function instance every time it's called, react sees that the new instance isn't `===` to the last one it saw, so it updates the DOM. We try to avoid doing that as much as possible for performance reasons (it causes unnecessary DOM updates and GC runs).

> A bind call or arrow function in a JSX prop will create a brand new function on every single render. This is bad for performance, as it will result in the garbage collector being invoked way more than is necessary. It may also cause unnecessary re-renders if a brand new function is passed as a prop to a component that uses reference equality check on the prop to determine if it should update.

*In this case, IMO we can leave it as-is*, because my experience tells me the performance hit won't be worth the refactor in this case. But if we were to change it, the correct way is typically to wrap the argument passing logic in the child component. So rather than this:

```javascript
<PillButton onClick={ this.onItemSelected(item) } />
```

you would do

```javascript
<PillButton 
  onClick={ this.onItemSelected }  // always sees the same function instance
  item={ item }
/>

// then in PillButton.js

handleClick = () => {
  // logic of what to pass to the callback goes here instead
  this.props.onClick(this.props.item)
}
```

(you can see that this isn't always optimal for making the code understandable. So we make considerations based on whether the performance hit will matter. but just FYI this is a pretty important concept.)

packages/veritone-react-common/src/components/Parts/Buttons/PillButton/index.js
olslash: appears not

packages/veritone-react-common/src/components/Parts/Buttons/PillButton/index.js
```diff
@@ -0,0 +1,60 @@
...
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+import defaultStyles from './styles.scss';
+
+export default class PillButton extends Component {
+	static propTypes = {
+		label: PropTypes.string,
+		info: PropTypes.string,
+		style: PropTypes.instanceOf(Object),
```
olslash: these should be `ObjectOf(any)`

packages/veritone-react-common/src/components/Parts/Buttons/PillButton/index.js
```diff
@@ -0,0 +1,60 @@
...
+		infoStyle: PropTypes.instanceOf(Object),
+		className: PropTypes.string,
+		gapClassName: PropTypes.string,
+		labelClassName: PropTypes.string,
+		infoClassName: PropTypes.string,
+		highlight: PropTypes.bool,
+		onClick: PropTypes.func
+	};
+
+	static defaultProps = {
```
olslash: doesn't look like these defaults are needed -- it's no problem to pass `undefined` in the case that they weren't provided

packages/veritone-react-common/src/components/Parts/Buttons/PillButton/index.js
```diff
@@ -0,0 +1,60 @@
...
+			gapClassName, 
+			labelClassName, 
+			infoClassName,
+			highlight,
+			onClick
+		} = this.props;
+		let hasGap = label && label.length > 0 && info && info.length > 0;
+
+		return (
+			<div className={highlight? classNames(defaultStyles.pillButton, defaultStyles.highlight, className) : classNames(defaultStyles.pillButton, className)} style={style} onClick={onClick}>
```
olslash: refactor to use `classnames` object style

packages/veritone-react-common/src/components/Parts/Buttons/PillButton/index.js
```diff
@@ -0,0 +1,60 @@
...
+				<div className={classNames(defaultStyles.label, labelClassName)} style={labelStyle}>
+					{label}
+				</div>
+				{
+					hasGap ? <div className={classNames(defaultStyles.gap, gapClassName)} style={gapStyle} /> : null
```
olslash: in JSX you do conditionals like this:

```javascript
{
  hasGap &&
  <div className={classNames(defaultStyles.gap, gapClassName)} style={gapStyle} /> 
}
```

packages/veritone-react-common/src/components/Parts/Buttons/PillButton/index.js
```diff
@@ -0,0 +1,60 @@
...
+import defaultStyles from './styles.scss';
+
+export default class PillButton extends Component {
+	static propTypes = {
+		label: PropTypes.string,
+		info: PropTypes.string,
+		style: PropTypes.instanceOf(Object),
+		gapStyle: PropTypes.instanceOf(Object),
+		labelStyle: PropTypes.instanceOf(Object),
+		infoStyle: PropTypes.instanceOf(Object),
```
olslash: is there a strong reason to support both overriding with both inline styles _and_ classnames? I'd prefer to not have inline styles at all if possible

packages/veritone-react-common/src/components/Parts/Buttons/PillButton/styles.scss
```diff
@@ -0,0 +1,42 @@
...
+  font-size: 14px;
+  display: inline-block;
+  background-color: white;
+
+  &:hover {
+    cursor: pointer;
+  }
+
+  > div {
+    float: left;
```
olslash: in modern HTML using float for positioning is a totally unnecessary hack -- you should be using flexbox instead.

KhuongVeritone:
We won't have the header in this component & the passed in data don't have any reference to the engine info anymore. So I'm closing this pull request, redo a bunch of stuff and create a new pull request later when it's done.

PR #69, opened 2018-03-23
use object assign eslint for vscode

charliemantle:


packages/veritone-redux-common/.eslintrc.js
```diff
@@ -1,22 +1,24 @@
...
-    'import'
-  ],
-  rules: {
-    ...env.eslintReact.rules,
-    'import/order': 2
-  }
-};
+module.exports = Object.assign({}, env.eslintReact, {
+  globals: Object.assign(
+    {
```
olslash: these need to be reversed, so `({}, globals, { local overrides })` to match the current behavior

idea is to be able to override globals in local configs

charliemantle: updated

PR #78, opened 2018-03-28
Feature/vtn 6758 logo recognition engine output component

KhuongVeritone:


packages/veritone-react-common/src/helpers/time.js
```diff
@@ -1,14 +1,37 @@
...
-  s = s % 60;
-  h = Math.floor(m / 60);
-  m = m % 60;
-
-  h = h < 10 && h > 0 ? '0' + h : h;
-  m = m < 10 ? '0' + m : m;
-  s = s < 10 ? '0' + s : s;
-
-  return (h > 0 ? h + ':' : '') + m + ':' + s;
+export function msToReadableString(
```
KhuongVeritone: We are using this because the date library in helpers doesn't return the values we wants

packages/veritone-react-common/src/components/Parts/Buttons/PillButton/index.js
```diff
@@ -0,0 +1,73 @@
+import React, { Component } from 'react';
+import { string, bool, func, objectOf, any } from 'prop-types';
+import classNames from 'classnames';
+import defaultStyles from './styles.scss';
+
+export default class PillButton extends Component {
+  static propTypes = {
+    label: string,
+    info: string,
+    style: objectOf(any),
```
KhuongVeritone: I don't have any strong reason why I have both style object & class name overwrite here. Just have extra time and want to provide both so that if someone wants to overwrite an existing style and doesn't want to use "!important"

AlexOVeritone: ok

packages/veritone-react-common/src/helpers/time.js
AlexOVeritone: Document the return format.

AlexOVeritone: Actually it looks good

PR #84, opened 2018-04-05
Fixed split table issue where focused table row was wrong value

gitrite:


packages/veritone-react-common/src/components/DataTable/story.js
```diff
@@ -169,22 +169,72 @@ function BasicTable({ data }) {
...
+    </Table>
   );
 }
 
 BasicTable.propTypes = {
   data: arrayOf(object)
 }
 
+class BasicSplitTable extends React.Component {
+  static propTypes ={
```
charliemantle: missing a space here, I would expect this to get caught by prettier.

PR #90, opened 2018-04-07
Media details page

AlexOVeritone:
React MDP with merged-in sdo branch


packages/veritone-react-common/src/components/MediaInfoPanel/EditMetadataDialog/index.js
```diff
@@ -0,0 +1,218 @@
...
+    metadata: objectOf(any),
+    isOpen: bool,
+    onSave: func,
+    onClose: func
+  };
+
+  state = {
+    filename: get(this.props.metadata, 'veritoneFile.filename', ''),
+    source: get(this.props.metadata, 'veritoneCustom.source', ''),
+    programLiveImage: get(this.props.metadata, 'veritoneProgram.programLiveImage', ''),
```
atb91590: is there an option to use the signed version of programLiveImage/programImage?
signed image to display and unsigned when saving?

btw, we should also pass the signed image into the media player to display

AlexOVeritone: yeah, will do.
Will take into account setting signed image to the player. I don't think we set image do this right now in this MDP implementation

AlexOVeritone: Btw just tried signed uri and got 403 access denied - missing some auth stuff.
I think it's ok to go as it is right now
I will fix this 403 for signed uri sometime this week

PR #100, opened 2018-04-17
Fix/table cell width

gitrite:
This PR moves the width prop from table cell into the style prop. This PR also implements the requested changes allowing for a scrollable table

### DIFF
--- a/packages/veritone-react-common/src/components/DataTable/index.js
+++ b/packages/veritone-react-common/src/components/DataTable/index.js
@@ -85,8 +85,8 @@ class _Table extends React.Component {
             key="split"
           />
         ) : (
-            <NormalTableContainer {...restProps} key="normal" />
-          )}
+          <NormalTableContainer {...restProps} key="normal" />
+        )}
       </CSSTransitionGroup>
     );
   }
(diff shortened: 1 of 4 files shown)


packages/veritone-react-common/src/components/DataTable/story.js
```diff
@@ -45,7 +45,12 @@ class BasicTable extends React.Component {
         showHeader
         onCellClick={this.handleCellClick}
       >
-        {columns.map(c => <Column key={c.dataKey} {...c} />)}
+        {columns.map((c, idx) => {
+          if (idx === columns.length - 1) {
+            c.width = 300;
```
olslash: don't mutate stuff. Also this should just go into the definition of columns.

```javascript

const columns = map(data[0], (val, key) => {
  return {
    dataKey: key,
    header: startCase(key),
    cellRenderer: flow([upperFirst, truncate]),
    align: 'center',
// here:
    width: key === 'ip' ? 300 : undefined
  };
});
```

packages/veritone-react-common/src/components/DataTable/index.js
```diff
@@ -319,12 +323,12 @@ export const Column = ({
     }
 
     return isFunction(cellRenderer)
-      ? cellRenderer(data[dataKey], data)
-      : String(data[dataKey] || '');
+      ? cellRenderer(get(data, dataKey), data)
```
olslash: seems like an unrelated change -- how come we need this now?

PR #101, opened 2018-04-18
Fds plus sdo integration

gitrite:


packages/veritone-react-common/package.json
```diff
@@ -41,7 +41,7 @@
     "prebuild": "yarn run clean:build",
     "build": "yarn run build:es && yarn run build:cjs",
     "build:es": "cross-env NODE_ENV=production BABEL_ENV=production-es rollup -c rollup.config.es.js",
-    "build:cjs": "cross-env NODE_ENV=production BABEL_ENV=production-cjs rollup -c rollup.config.cjs.js",
+    "build:cjs": "cross-env NODE_ENV=production BABEL_ENV=production-cjs rollup -c rollup.config.cjs.js --watch",
```
atb91590: don't think we should include this since it does collide with the buildall script in the root dir.

PR #116, opened 2018-04-25
Feature/vtn 7342 mdp integrate structured data

AlexOVeritone:
Add StructuredData/Corelation engine output.
NOTE: does not have NoData and Fetching states yet.


packages/veritone-react-common/src/components/IngestionAdapters/DynamicAdapter/index.js
```diff
@@ -102,7 +101,9 @@ class DynamicAdapter extends React.Component {
             </div>
           </div>
           <div>
-            <div className={styles.adapterContainer + ' ' + styles.flexContainer}>
+            <div
+              className={styles.adapterContainer + ' ' + styles.flexContainer}
```
ChrisPelletier: use https://www.npmjs.com/package/classnames you can "import classnames as cx from 'classnames'' then do replace this with "cx(styles.adapterContainer, styles.flexContainer)"

packages/veritone-react-common/src/components/StructuredDataEngineOutput/story.js
```diff
@@ -0,0 +1,176 @@
...
+  let mockSchemaById = getMockSchemaById();
+
+  return (
+    <StructuredDataEngineOutput
+      className={styles.outputViewRoot}
+      data={mockEngineOutput}
+      schemaById={mockSchemaById}
+      engines={engines}
+      selectedEngineId={'engineOfInterest'}
+      onEngineChange={onEngineSelected}
```
ChrisPelletier: I think the correct way to do something like this is action knobs https://github.com/storybooks/storybook/tree/master/addons/actions

packages/veritone-react-common/src/components/StructuredDataEngineOutput/index.js
```diff
@@ -0,0 +1,198 @@
...
+    this.setState({
+      selectedSchemaId: null,
+      flattenStructuredData: {},
+      engineSchemaIds: []
+    });
+    this.props.onEngineChange(engineId);
+  };
+
+  render() {
+    let {
```
KhuongVeritone: should this be const instead of let?

packages/veritone-react-common/src/components/StructuredDataEngineOutput/index.js
```diff
@@ -0,0 +1,198 @@
...
+      className,
+      engines,
+      schemaById,
+      selectedEngineId,
+      onExpandClicked
+    } = this.props;
+
+    let {
```
KhuongVeritone: should this be const instead of let?

PR #119, opened 2018-04-27
VTN-7173 add a default ISO list of languages

KhuongVeritone:
I added a simple list of default languages base on ISO & IETF standards. I'm going to move this into a separate package and expand it to a full list 5 times bigger than this. However, that will take some time with a lot of copy & paste. This PR helps to stop unblocking the TranslationEngineOutput Component for now.


charliemantle:
How comes there is a need to create our own package, isn't there something available we can use that we don't need to maintain?

charliemantle:
The translation layer is to be expected, but I disagree with including all of the language codes / regexes in the SDK for us to maintain, rather than using something that already exists. For example why wouldn't something like https://github.com/georgzoeller/iso-3166-1-alpha-2 be suitable?

KhuongVeritone:
Yeah, I looked into a few of them. And I couldn't find any good one. Like the one you just mentioned, if you looked into their code you can see that it's just a copy list from Wikipedia so some of the names are messed up.

Ex:
BO: "Bolivia, Plurinational State of",
BQ: "Bonaire, Sint Eustatius and Saba", 
VA: "Holy See (Vatican City State)" where the official name is Vatican City
... 

I'm a bit skeptical when using free libs from random people. Always find something here and there if you dive to their code.

packages/veritone-widgets/src/widgets/MediaDetails/index.js
```diff
@@ -684,15 +684,14 @@ class MediaDetailsWidget extends React.Component {
                       <TranslationEngineOutput
                         contents={engineResultsByEngineId[selectedEngineId]}
                         onClick={this.handleUpdateMediaPlayerTime}
-                        //onRerunProcess={handle rerun process callback}
+                        //onRerunProcess={handle rerun process callback if needed}
```
AlexOVeritone: This one you can pass already

packages/veritone-widgets/src/widgets/MediaDetails/index.js
```diff
@@ -684,15 +684,14 @@ class MediaDetailsWidget extends React.Component {
...
                         className={styles.engineOuputContainer}
                         engines={selectedEngineCategory.engines}
                         selectedEngineId={selectedEngineId}
                         onEngineChange={this.handleSelectEngine}
                         onExpandClicked={this.toggleExpandedMode}
-                        //languages={language options}
                         //defaultLanguage={'en-US'}
```
AlexOVeritone: remove commented code.
onLanguageChange should not be handled outside

KhuongVeritone:
Actually, the last one won't work either since it won't cover cases like 'American English' or 'British English'.

olslash:
if you want to make one for our uses, I'd just ask that you make a new repo in the veritone org and put it there -- not part of the SDK

charliemantle:
@KhuongVeritone thank you for being thorough with this, really great to see good job. 👍

PR #136, opened 2018-05-12
VTN-5961 & VTN-5985 handle image media type in MDP

KhuongVeritone:


packages/veritone-widgets/src/widgets/MediaDetails/index.js
```diff
@@ -598,12 +601,21 @@ class MediaDetailsWidget extends React.Component {
...
-                  />
+                  {isImage ? (
+                    <Image
+                      src={this.getPrimaryAssetUri()}
+                      width="100%"
+                      height="100%"
+                      type="contain"
+                    />
+                  ) : (
+                    <MediaPlayer
```
atb91590: @SteveShaffer should we keep image support functionality separate from the media player itself? the implementation in the angular media player component was done within it (to prevent repetitive if-image conditions being scattered thru-out apps where ever the player is used).

if we want the have the react media player to support images (or any future asset types), we'll need to pass the mimetype in as a parameter. an example of a potential future asset type is .pdf - we can possibly extend the media player to display pdfs...?

PR #137, opened 2018-05-15
This enables the login with only a cookie after they try the login me…

mclenithan:
…thod


packages/veritone-redux-common/src/modules/user/index.js
```diff
@@ -176,7 +176,8 @@ export function fetchUser() {
       ],
       endpoint: state => `${getConfig(state).apiRoot}/v1/admin/current-user`,
       method: 'GET',
-      headers: commonHeaders
+      headers: commonHeaders,
+      credentials: 'include'
```
mclenithan: This way, the login method must be called, which calls `fetchUser()`, and the bearer token will be set if the user has a cookie.  This will not enable all API endpoints to try to include credentials, which can change if this is not what we want.

PR #142, opened 2018-05-21
VTN-6038: Edit Profile Component

bgilkeyveritone:
Seeming how this is my first react component, i would like to get a CR on this now on initial completion. still waiting on the smart file picker component, and still need to write test, which i will do once i know i did this correctly, and have the code in a completed state. 

Will be talking with PO on this so i can remove some of the style overrides as well.


charliemantle:
While you're waiting for this to be reviewed you should install eslint / prettier plugins and fix the errors/warnings as it looks like you don't have the plugins enabled. Also take a look at some of the other css files and see how we're using color variables / mui typography as this is something you should try to adhere to.

charliemantle:
no explicit changes, just a note to ensure you're using eslint / prettier.

PR #148, opened 2018-05-29
add tests for image component

atb91590:
https://steel-ventures.atlassian.net/browse/VTN-8057

### DIFF
--- a/packages/veritone-react-common/src/components/Image/index.test.js
+++ b/packages/veritone-react-common/src/components/Image/index.test.js
@@ -27,7 +27,7 @@ describe('Image', () => {
   });
 
   it('Border can be set', () => {
-    const wrapper = mount(<Image src={TEST_IMG} border={true} />);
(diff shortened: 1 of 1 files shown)


packages/veritone-react-common/src/components/Image/index.test.js
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+import { mount } from 'enzyme';
+
+import Image from './';
+
+describe('Image', () => {
+  const TEST_IMG = 'https://pp.userapi.com/c624631/v624631866/38602/miSZV1RlzEY.jpg';
+  
+  it('Image should render in a div tag', () => {
```
charliemantle: can you modify these to conform to a sentence, i.e 

`it('should render the image in a div tag')`
`it('should have a default height/width')`
`it('allows for height/width to be specified')`
`it('allows for a border to be set')`

atb91590: yep, i sure can

PR #158, opened 2018-06-04
Expandable search field for saved cognitive profiles

atungw:


packages/veritone-react-common/src/components/ExpandableSearchField/index.js
```diff
@@ -0,0 +1,105 @@
...
+  };
+
+  onKeyPress = evt => {
+    if (
+      evt.key === 'Enter' &&
+      this.props.onSearch &&
+      evt.target.value &&
+      evt.target.value.trim().length > 0
+    ) {
+      this.props.onSearch(evt.target.value.trim());
```
olslash: pretty sure trimming should be done by the _consumer_ of the search, since there may be a case where whitespace is important. More generic that way, and consistent with other html inputs.

PR #162, opened 2018-06-11
Show redactionStatus tag and readonly custom tags.

AlexOVeritone:


packages/veritone-react-common/src/components/MediaInfoPanel/EditTagsDialog/tagPill/index.js
```diff
@@ -16,8 +16,15 @@ class TagPill extends Component {
   };
 
   render() {
+    console.log('render pill', this.props.text);
```
ChrisPelletier: remove this console log.

PR #164, opened 2018-06-12
Feature/vtn 8000 mdp context menu

atb91590:
these changes are to keep the menu on the MediaInfoPanel component for editing metadata/tags, but moving all other overflow menu items over to MediaDetails component (download & context menu items).


packages/veritone-widgets/src/widgets/MediaDetails/index.js
```diff
@@ -347,6 +343,10 @@ class MediaDetailsWidget extends React.Component {
     return !this.isMediaPublic() && this.isOwnMedia();
```
AlexOVeritone: Could you do here
return this.canEditMedia();

PR #165, opened 2018-06-12
Sdo integration transcript/face save dialogs

ChrisPelletier:
In certain scenarios when a user is editing a transcript or face engine output we need to warn them that their edits will be removed if they perform certain actions.

If the user has made edits and they hit the cancel or back button it should display a modal that asks them if they want to save their changes.

If they are in face or transcript and they change the engine id we need to warn them that their changes will be lost.

In transcript if they change the edit mode it should warn them that their changes will be lost.

### DIFF
--- a/packages/veritone-widgets/src/redux/modules/mediaDetails/saga.js
+++ b/packages/veritone-widgets/src/redux/modules/mediaDetails/saga.js
@@ -68,6 +68,7 @@ import {
   isEditModeEnabled,
   isUserGeneratedTranscriptEngineId,
   isUserGeneratedFaceEngineId,
+  toggleEditMode,
   getSelectedEngineCategory,
   refreshEngineRunsSuccess,
   clearEngineResultsByEngineId,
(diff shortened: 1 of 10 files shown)


packages/veritone-widgets/src/redux/modules/mediaDetails/saga.js
```diff
@@ -75,7 +79,9 @@ import {
   getEngineCategories
 } from '.';
 
-import { ADD_DETECTED_FACE } from './faceEngineOutput';
+import {
+  ADD_DETECTED_FACE
```
gitrite: this should be included with the previous import declaration from faceEngineOutput

packages/veritone-widgets/src/widgets/TranscriptEngineOutputWidget/index.js
```diff
@@ -137,24 +140,35 @@ export default class TranscriptEngineOutputWidget extends Component {
...
-          pendingEditMode: undefined
-        };
-      });
+      this.props.onEngineChange(engineId);
     }
   };
 
+  handleAlertConfirm = () => {
+    this.props.reset();
+    this.state.alertConfirmAction();
```
gitrite: this seems odd. shouldn't the function be class method?

packages/veritone-widgets/src/widgets/TranscriptEngineOutputWidget/index.js
```diff
@@ -137,24 +140,35 @@ export default class TranscriptEngineOutputWidget extends Component {
     }
   };
 
-  handleAlertConfirm = value => {
-    if (value === 'cancel') {
+  handleEngineChange = engineId => {
+    if (this.props.editMode && this.props.hasChanged) {
```
gitrite: probably should use a better naming that `hasChanged`?

PR #168, opened 2018-06-13
Feature/vtn 8060 clean up transcript engine output

KhuongVeritone:
Fixed switching engine bug passing engine ID to child components.
Added unit tests
The rest are just yarn format


KhuongVeritone:
Tested with 400003774 & 400004914. And it seems to work properly for me

PR #172, opened 2018-06-20
Integrate media player

olslash:
This brings over the MediaPlayer and Video stuff from the sdo-integration branch @JordanBelford


JordanBelford:
@olslash On the sdo-integration branch, we also included `video-react` and `shaka-player` as external in the rollup config. https://github.com/veritone/veritone-sdk/blob/sdo-integration/packages/veritone-react-common/rollup.config.base.js#L42
Other than that looks good

olslash:
true, thanks, will add that

PR #179, opened 2018-06-27
Feature/vtn 8648 source type field

gitrite:
https://steel-ventures.atlassian.net/browse/VTN-8648


olslash:
https://github.com/veritone/veritone-sdk/pull/186

PR #185, opened 2018-06-29
Added SDOTable component and widget

gitrite:
https://steel-ventures.atlassian.net/browse/VTN-8722


packages/veritone-react-common/src/components/SDOTable/index.js
```diff
@@ -0,0 +1,69 @@
...
+        formattedData = data;
+    }
+
+    return <DotDotDot clamp={2}>{formattedData}</DotDotDot>;
+  };
+
+  render() {
+    const tableProps = omit(this.props, ['data', 'schema', 'paginate']);
+
+    const TableComp = this.props.paginate ? PaginatedTable : Table;
```
charliemantle: really minor but would prefer to name this TableComponent, not gaining anything by shortening the variable here.

packages/veritone-widgets/src/widgets/SDOTable/story.js
```diff
@@ -0,0 +1,120 @@
+import React from 'react';
```
charliemantle: This should be refactored to use the BaseStory to abstract some of the boilerplate

PR #188, opened 2018-07-09
Feature/vtn 8313 edit transcript

ChrisPelletier:
https://steel-ventures.atlassian.net/browse/VTN-8313

Updated the fetch engine results graphql query to accept an ignoreUserEdited flag that tells graphql to return user edited engine results or not. This is needed for engine categories that allow editing to display the user edited or original engine results. I also added a material-ui Select that should only display if there are user edited results and allows the user to toggle between the user edited results and original.


packages/veritone-widgets/src/redux/modules/mediaDetails/faceEngineOutput/index.js
```diff
@@ -546,24 +555,26 @@ export const getFaceEngineAssetData = (state, engineId) => {
...
+        ...engineResult
+      }
+    ];
   }, []);
 
-  return {
-    ...userEdited,
-    series: allSeries
-  };
+  return allJsonData;
```
AlexOVeritone: as previously this should collapse all series into one array, preferably sort it, and return single JSON, not array

packages/veritone-widgets/src/redux/modules/mediaDetails/saga.js
```diff
@@ -679,23 +685,38 @@ function* createFileAssetSaga(
   sourceData,
   fileData
 ) {
+  if (sourceData.sourceEngineId) {
```
AlexOVeritone: Move this to watchSaveAssetData before yielding to createFileAssetSaga.
createFileAssetSaga should be more generic. We use it to create plain/text asset for bulk transcript edit.

PR #196, opened 2018-07-17
Engine Selection - added separate search and filter states for tabs + UX improvements

charliemantle:
- Exported `DiscoverySideBarContainerPure`
- Added separate search and filter states for each tab
- Added redux logic for fetching / handling engine categories
- Added async loading of engine categories to the category filter
- Made engine detail page a full page takeover
- Now clears all filters on tab switch
- Now resets filter active path on tab switch
- Removed categories filter title
- Changed default color of lozenge (push/pull pills)
- Added border between sidebar and content
- Filters now work on ‘Own Engines’ Page
- Added count around Your Engines Tab
- Search now works on both tabs
- Selected filters now persist on each tab
- Added new state when on "Own Engines" tab when there no filter results

### DIFF
--- a/packages/veritone-react-common/src/components/Lozenge/index.js
+++ b/packages/veritone-react-common/src/components/Lozenge/index.js
@@ -1,6 +1,6 @@
 import React from 'react';
 import { string } from 'prop-types';
-import blue from '@material-ui/core/colors/blue';
+import indigo from '@material-ui/core/colors/indigo';
 
 import styles from './styles.scss';
 
@@ -9,7 +9,7 @@ const Lozenge = ({ children, iconClassName, backgroundColor, textColor }) => {
     <div
       className={styles.lozenge}
       style={{
-        backgroundColor: backgroundColor || blue[500],
+        backgroundColor: backgroundColor || indigo[500],
         color: textColor || '#fff'
       }}
     >
--- a/packages/veritone-widgets/src/redux/modules/engineSelection/index.js
+++ b/packages/veritone-widgets/src/redux/modules/engineSelection/index.js
@@ -1,4 +1,15 @@
-import { get, set, omit, without, union, merge, difference } from 'lodash';
+import {
+  get,
+  set,
+  omit,
+  without,
+  union,
+  merge,
+  difference,
+  isNumber,
+  isEmpty,
+  intersection
+} from 'lodash';
 import { helpers, modules } from 'veritone-redux-common';
 const { createReducer } = helpers;
 const { engine: engineModule } = modules;
@@ -24,6 +35,9 @@ export const CLEAR_SEARCH = 'vtn/engineSelection/CLEAR_SEARCH';
 export const CHANGE_TAB = 'vtn/engineSelection/CHANGE_TAB';
 export const TOGGLE_SEARCH = 'vtn/engineSelection/TOGGLE_SEARCH';
 
+export const SET_ACTIVE_SIDEBAR_PATH =
+  'vtn/engineSelection/SET_ACTIVE_SIDEBAR_PATH';
+
 export const SET_DESELECTED_ENGINES =
   'vtn/engineSelection/SET_DESELECTED_ENGINES';
 export const SET_ALL_ENGINES_SELECTED =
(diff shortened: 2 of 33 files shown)


packages/veritone-widgets/src/redux/modules/engineSelection/index.js
```diff
@@ -294,7 +375,7 @@ export function initializeWidget(id) {
   };
 }
 
-export function refetchEngines(id) {
+export function refetchEngines(id, currentTabIndex) {
```
olslash: why does this need an ID? relatedly, enginesModule.fetchEngines doesn't even use the ID you have it accepting...

packages/veritone-widgets/src/widgets/EngineSelection/EngineListView/SelectBar/SearchBar.js
```diff
@@ -25,6 +26,24 @@ export default class SearchBar extends React.Component {
     searchQuery: ''
   };
 
+  state = {
+    searchQuery: this.props.searchQuery
+  };
+
+  static getDerivedStateFromProps(nextProps, prevState) {
```
olslash: this is sketchy pls read https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html and reconsider

packages/veritone-widgets/src/widgets/EngineSelection/EngineListView/SelectBar/SearchBar.js
```diff
@@ -39,9 +58,10 @@ export default class SearchBar extends React.Component {
   };
 
   handleChange = event => {
-    if (event.target.value === this.props.searchQuery) {
-      return;
-    }
+    this.setState({
```
olslash: do you really need this state? why can't this be a normal controlled component?

packages/veritone-widgets/src/widgets/EngineSelection/EngineListView/SelectBar/index.js
```diff
@@ -31,6 +32,10 @@ export default class SelectBar extends React.Component {
   };
 
   render() {
+    if (this.props.hideActionMenuItems) {
```
olslash: seems like "hide action menu items" actually means "hide this entire component", in which case, this hiding logic should be in the parent component that renders this one.

packages/veritone-widgets/src/widgets/EngineSelection/EngineListView/SelectBar/index.js
```diff
@@ -41,46 +46,45 @@ export default class SelectBar extends React.Component {
...
+            onToggleSearch={this.props.onToggleSearch}
+            searchQuery={this.props.searchQuery}
+            isOpen={this.props.isSearchOpen}
+          />
+          {this.props.actionMenuItems && (
+            <div className={styles.actionMenuDividerContainer}>
+              <div className={styles.actionMenuDivider} />
+            </div>
+          )}
+          {this.props.actionMenuItems &&
```
olslash: try not to use `return` in these inner render functions (i wrote about this in the react standards doc)

also you shouldn't use index as a key... button text + icon class should be unique among items, right?

```javascript
{this.props.actionMenuItems &&
            this.props.actionMenuItems.map(
              ({ buttonText, iconClass, onClick }) =>
                buttonText ? (
                  <Button
                    key={`${buttonText}-${iconClass}`}
                    onClick={onClick}
                    color="primary"
                  >
                    {buttonText}
                  </Button>
                ) : (
                  <IconButton
                    key={`${buttonText}-${iconClass}`}
                    onClick={onClick}
                  >
                    <i className={iconClass} />
                  </IconButton>
                )
            )}
```

packages/veritone-react-common/src/components/Lozenge/index.js
```diff
@@ -9,7 +9,7 @@ const Lozenge = ({ children, iconClassName, backgroundColor, textColor }) => {
     <div
       className={styles.lozenge}
       style={{
-        backgroundColor: backgroundColor || blue[500],
+        backgroundColor: backgroundColor || indigo[500],
```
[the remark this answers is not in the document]
charliemantle: I made this change based on the feedback that it looks too much like a button using the default mui color.

packages/veritone-widgets/src/redux/modules/engineSelection/index.js
charliemantle: you're right, I added this by mistake, will fix.

PR #198, opened 2018-07-18
Feature/vtn 8768 add restore original menu

AlexOVeritone:
Link to Ticket: https://steel-ventures.atlassian.net/browse/VTN-8768

To merge into the branch this one originated from.

### DIFF
--- a/packages/veritone-react-common/src/components/EngineOutputHeader/index.js
+++ b/packages/veritone-react-common/src/components/EngineOutputHeader/index.js
@@ -64,7 +64,7 @@ class EngineOutputHeader extends Component {
   toggleIsMoreMenuOpen = () => {
     this.setState(prevState => {
       return {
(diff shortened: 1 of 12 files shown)


packages/veritone-react-common/src/components/EngineOutputHeader/index.js
```diff
@@ -46,12 +61,86 @@ class EngineOutputHeader extends Component {
     }
   };
 
+  toggleIsMoreMenuOpen = () => {
+    this.setState(prevState => {
+      return {
+        isMoreMenuOpen: !{ ...prevState }.isMoreMenuOpen
```
ChrisPelletier: you shouldn't need to destructure prevState. You should just be able to do { isMoreMenuOpen: !prevState .isMoreMenuOpen }

AlexOVeritone: done

PR #200, opened 2018-07-20
Do not hardcode z-index in the FullScreenDialog

AlexOVeritone:
instead, allow component users to set it.
Run Process dialog has z-index:80, MDP - 50. SourceManagementForm - 950.

### DIFF
--- a/packages/veritone-react-common/src/components/FullScreenDialog/styles.scss
+++ b/packages/veritone-react-common/src/components/FullScreenDialog/styles.scss
@@ -4,7 +4,6 @@
   left: 0;
   right: 0;
   bottom: 0;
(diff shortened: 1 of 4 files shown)


olslash:
this'll break any other places that use FullScreenDialog and expect the z-index to be set by default -- can you just make local changes override the default? (maybe move the classname [here](https://github.com/veritone/veritone-sdk/blob/fa325237120351dd7479938a238cb80b37d3da5c/packages/veritone-react-common/src/components/FullScreenDialog/index.js#L19) up above the others?

olslash:
remember that it's also an exported component, so this would be a breaking change (major version # bump). for example, developer app uses it.

PR #201, opened 2018-07-25
Updates to allow for custom themes

atungw:
- Add themeing parameter to withMuiThemeProvider that merges in any themes provided. (No longer used, but provided as a decorator for people using veritone-react-common directly).
- Add namespace and a JSS provider to prevent collisions with other material-ui components
- Remove withMuiThemeProvider from components and add a passthrough provider for storybook
- Add theming provider to the root of VeritoneApp, configurable with VeritoneApp({theme: ANY_THEME}).
- Remove theme from engineSelection widget.

Unsure how `veritone-react-common` components should get their styles. Right now we decorate everything with withMuiProvider, but it's less than ideal because you're technically only supposed to use MuiThemeProvider once at the root of your tree (for performance reasons). https://material-ui.com/api/mui-theme-provider/

We also definitely can't use a style wrapped `veritone-react-common` in a widget, since it'll basically either prevent the custom theme from being passed on (which should only be defined in one place anyways for sanity), or we merge in their themes into ours, which I think is bad since it'll be surprising and indeterminate. 

I'm thinking our best bet at the moment is to just export the HOC component that will set our own theme as well as the decorator, and put a note to include it at the top of the rendering tree if they're not using and material-ui / JSS, or to decorate any uses of the `veritone-react-common` components if they are.

See: https://material-ui.com/getting-started/faq/ `How to fix a class names conflict in production?` for why we need JssProvider.

This is testable by doing something like: `new VeritoneApp({theme: { typography: { htmlFontSize: 10 } } })` which will make any component rendered use larger font sizes.


packages/veritone-react-common/src/helpers/withMuiThemeProvider.js
```diff
@@ -1,48 +1,68 @@
...
+  MuiThemeProvider,
+  createMuiTheme,
+  jssPreset
+} from '@material-ui/core/styles';
+import { create } from 'jss';
+import JssProvider from 'react-jss/lib/JssProvider';
+
+import { guid } from './guid';
+
+export class VSDKStyleWrapper extends React.PureComponent {
```
olslash: believe PureComponent isn't suitable for a top level component because it'll prevent children from updating properly: https://reactjs.org/docs/react-api.html#reactpurecomponent

> Furthermore, React.PureComponent’s shouldComponentUpdate() skips prop updates for the whole component subtree. Make sure all the children components are also “pure”.

packages/veritone-react-common/src/helpers/withMuiThemeProvider.js
```diff
@@ -1,48 +1,68 @@
...
+  constructor() {
+    super();
+    this._style = `vsdk-${guid()}`;
+    this._jss = create(jssPreset());
   }
 
-  function forwardRef(props, ref) {
-    return <WrappedWithMuiTheme {...props} forwardedRef={ref} />;
+  render() {
+    const mergedTheme = this.props.theme
```
olslash: since props.theme will be undefined, you don't need the conditional (also the cloning step is unnecessary). so in all cases

```javascript
 const mergedTheme = createMuiTheme(this.props.theme)
```
should be fine.

also what happened to the default palette? I think we should indeed deep-merge user themes with ours (with the passed-in theme as higher priority).

packages/veritone-react-common/.storybook/config.js
```diff
@@ -1,14 +1,22 @@
-import React from 'react';
+import React, {Fragment} from 'react';
```
olslash: unused

packages/veritone-react-common/.storybook/config.js
```diff
@@ -1,14 +1,22 @@
...
 
 function loadStories() {
   req.keys().forEach(filename => req(filename));
 }
 
+const withStyles = (story) => (
+  <VSDKStyleWrapper>
+    { story() }
+  </VSDKStyleWrapper>
+)
```
olslash: this would have a semi if prettier was run, please make sure to run it over the whole feature

PR #210, opened 2018-08-07
VTN-9167 export engine output smart component/widget

ChrisPelletier:
A new smart component/widget that will allow a user to configure an engineOutputExport configuration per task 2.4 in the following tech spec (https://steel-ventures.atlassian.net/wiki/spaces/VT/pages/728596944/Bloomberg+Captioning+Technical+Specifications#BloombergCaptioningTechnicalSpecifications-Task2.4:Createdownloadtransformedengineoutputcomponent) It will accept an array of tdo objects that specify the tdo and optionally a start and stop offset. It then determines the engines ran on the tdos and sets up the initial configuration. The user can then perform different actions to configure it further.


packages/veritone-widgets/src/redux/modules/engineOutputExport/saga.js
```diff
@@ -0,0 +1,222 @@
...
+  getTdoData,
+  exportAndDownloadFailure,
+  addSnackBar,
+  storeSubtitleConfigs
+} from './';
+import { guid } from '../../../shared/util';
+
+const { auth: authModule, config: configModule } = modules;
+
+function* fetchEngineRuns(tdoIds) {
```
olslash: we shouldn't be writing these fetching actions as sagas -- constantly rewriting the fetch logic opens us up to a lot of bugs. For example this doesn't work with oauth because you forgot to add the oauth token.

use the `callGraphQLApi` from `veritone-redux-common` from a thunk in your module instead (example: https://github.com/veritone/veritone-sdk/blob/c7af53286b4df4ce8d54fb9a1095da26dbebe29f/packages/veritone-redux-common/src/modules/engine/index.js#L181)

PR #220, opened 2018-08-20
Integrate export engine output into mdp

ChrisPelletier:
Integrated the EngineOutputExport component into MDP and updated the EngineOutputHeader component to accept an array of nodes/components rather than an array of objects. I also added a ExportMenuItem component since it is a little more complex than a regular menu item with it's own functionality and submenu. I also added the createQuickExport function for creating an export without having to do a detailed configuration with the EngineOutputExport component. I renamed the TranscriptEngineOutputWidget to TranscriptEngineOutputContainer since it isn't actually a widget.


packages/veritone-widgets/src/widgets/MediaDetails/index.js
```diff
@@ -808,6 +864,33 @@ class MediaDetailsWidget extends React.Component {
...
 
+    const moreMenuItems = [];
+    const selectedEngine = find(get(selectedEngineCategory, 'engines', []), {
+      id: selectedEngineId
+    });
+    if (!isEditModeEnabled && get(selectedEngine, 'hasUserEdits')) {
+      moreMenuItems.push(
+        <MenuItem
+          key="restore-original-output"
+          className={moreMenuStyles.exportMainMenuItem}
```
AlexOVeritone: call this moreMenuItem

packages/veritone-widgets/src/widgets/MediaDetails/moreMenuItems/ExportMenuItem.js
```diff
@@ -0,0 +1,122 @@
+import React, { Component } from 'react';
```
AlexOVeritone: Keep the package name CamelCased

PR #232, opened 2018-08-27
Add SharingConfiguration that SourceManagementForm depends on.

AlexOVeritone:
When creating an ingestion job user should be able to set sharing options / acls on the source and/or program. ACL will be set with default permission level passed down to the component. For future component can be extended to include permission level selector, and default to the default permission value.


AlexOVeritone:
Postponing merge. Need to make substantial changes.

PR #241, opened 2018-09-06
ignore stylelint rule that conflicts with prettier

JordanBelford:
disables `value-list-comma-newline-after` because it conflicts with prettier


olslash:
where is this coming up? never seen it before

olslash:
example from chat:
```
box-shadow: 0 3px 5px -1px rgba(0, 0, 0, 0.2), 0 5px 8px 0 rgba(0, 0, 0, 0.14),
    0 1px 14px 0 rgba(0, 0, 0, 0.12);
```

prettier wraps because of line length limit, then style lint wants line breaks after every comma.

PR #254, opened 2018-09-28
Feature/vtn 9351 merge dynamic adapter

atb91590:
Ticket: https://steel-ventures.atlassian.net/browse/VTN-9351

These components were copied over from sdo-integration branch as of 2018.09.28

### DIFF
--- a/packages/veritone-react-common/src/components/InfiniteDropdownMenu/index.js
+++ b/packages/veritone-react-common/src/components/InfiniteDropdownMenu/index.js
@@ -56,6 +56,9 @@ export default class InfiniteDropdownMenu extends React.Component {
   };
 
   static defaultProps = {
+    loadNextPage: ({ startIndex, stopIndex }) => {
+      return Promise.resolve([]);
+    },
     pageSize: 30
   };
 
@@ -63,7 +66,7 @@ export default class InfiniteDropdownMenu extends React.Component {
     anchorEl: null
   };
 
-  UNSAFE_componentWillMount() {
+  componentDidMount() {
     if (isFunction(this.props.loadNextPage)) {
       this.props.loadNextPage({
         startIndex: 0,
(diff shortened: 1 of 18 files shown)


packages/veritone-react-common/src/components/InfiniteDropdownMenu/index.js
```diff
@@ -0,0 +1,313 @@
...
+
+  static defaultProps = {
+    pageSize: 30
+  };
+
+  state = {
+    anchorEl: null
+  };
+
+  UNSAFE_componentWillMount() {
```
KhuongVeritone: Can we move this to **componentDidMount** to avoid **UNSAFE**

packages/veritone-react-common/src/components/InfiniteDropdownMenu/index.js
```diff
@@ -0,0 +1,313 @@
...
+      </div>
+    );
+  };
+
+  openCustomTrigger = trigger => () => {
+    this.setState({ anchorEl: null }, trigger);
+  };
+
+  // To suppress InfiniteLoader warnings
+  dummyLoadNextPage = ({ startIndex, stopIndex }) => {
```
KhuongVeritone: I wonder if we can use this as the default value for loadNextPage

olslash:
this is broken and being reverted so it doesn't hold up other urgent releases

DynamicAdapter storybook broken:
image

Image storybook throwing proptype errors (use numbers for pixel sizes, not strigs)
image

InfiniteDropdownMenu is not in index.js

bunch of exports from DynamicAdapter that aren't components? react-common isn't a container for business logic. let's figure out a better way to deal with these helpers.

widgets don't follow documented package export style (look at first paragraph of readme, and at all other widgets as examples)

PR #265, opened 2018-10-11
Feature/vtn 11043 range slider

KhuongVeritone:
**Range Slider Component**
- Dual Range Slider
- Single Range Slider

[Jira Ticket: VTN-11043](https://steel-ventures.atlassian.net/browse/VTN-11043)

### DIFF
--- a/packages/veritone-react-common/src/components/RangeSlider/index.js
+++ b/packages/veritone-react-common/src/components/RangeSlider/index.js
@@ -4,6 +4,9 @@ import cx from 'classnames';
 
 import styles from './styles.scss';
 
+const lowerButton = 'lowerButton';
+const upperButton = 'upperButton';
+
 export default class RangeSlider extends React.Component {
   static propTypes = {
     min: number,
@@ -138,8 +141,6 @@ export default class RangeSlider extends React.Component {
     return newState;
   }
 
-  lowerButton = 'lowerButton';
-  upperButton = 'upperButton';
   handleMouseDown = event => {
     document.onmousemove = this.handleMouseMove;
     document.onmouseup = this.handleMouseUp;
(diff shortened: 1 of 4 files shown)


packages/veritone-react-common/src/components/RangeSlider/index.js
```diff
@@ -0,0 +1,367 @@
...
+      prevProps &&
+      min === prevProps.min &&
+      max === prevProps.max &&
+      value === prevProps.value &&
+      lowerValue === prevProps.lowerValue &&
+      upperValue === prevProps.upperValue &&
+      prefix === prevProps.prefix &&
+      suffix === prevProps.suffix &&
+      showValues === prevProps.showValues &&
+      numDecimalDigits === prevProps.numDecimalDigits
```
olslash: `isEqual(props, prevProps)`?

packages/veritone-react-common/src/components/RangeSlider/index.js
```diff
@@ -0,0 +1,367 @@
...
+    newState.upperValue = dual ? upperValue : value;
+    newState.initialLowerButtonVal = newState.lowerRatio;
+    newState.initialUpperButtonVal = newState.upperRatio;
+
+    return newState;
+  }
+
+  handleMouseDown = event => {
+    document.onmousemove = this.handleMouseMove;
+    document.onmouseup = this.handleMouseUp;
```
olslash: sketchy -- what if two components in an app wanted to both use document listeners like this? They'd be constantly clobbering each other. 

You should only be dealing with listeners within React itself (as props on an `<Element>`). This would probably be better done using html draggable/ondrag than global listeners?

KhuongVeritone:
hmm I wasn't aware of the double range slider in https://github.com/airbnb/rheostat. I'll try to use that to see if it works with what we want. Canceling this Pull Request for now

PR #275, opened 2018-10-24
VTN-11574 dynamically run widget sagas (again)

olslash:
this was previously merged but reverted https://github.com/veritone/veritone-sdk/pull/248

in this PR i forgot to account for storybook, and I didn't document the new requirement that smart component sagas be included in user-apps root sagas.

I also didn’t account for dependencies such as when a widget uses a smart component that has a saga.

More thought is needed.

Add a thing in BaseStory that starts/stops sagas, and add the requirement to README.md that if you use a smart component, you have to add its saga to your app's root saga.

also need to export the sagas if they aren't already.

it should probably be a major version because apps need to remove existing smart component sagas from their own root saga.


MmtBkn:
Closing as no activity over a long period of time.

PR #279, opened 2018-10-29
VTN-11759 after some refactoring processing fetch entites response go…

AlexOVeritone:
…t broken, fix it up and use new query.

Link to Ticket: https://steel-ventures.atlassian.net/browse/VTN-11759

Reviewers
@ChrisPelletier

Purpose
Fingerprint wuld request entities but response would not get stores in the redux.

What Changed
Store entities response in common redux, not per widget. Use new query to fetch entites

How should we monitor this?
Media details page, FigerPring output shout show entities and libraries.

### DIFF
--- a/packages/veritone-widgets/src/redux/modules/mediaDetails/saga.js
+++ b/packages/veritone-widgets/src/redux/modules/mediaDetails/saga.js
@@ -619,11 +619,12 @@ function* watchUpdateTdoContentTemplates() {
   });
 }
 
-function* fetchEntities(widgetId, entityIds) {
-  yield put({ type: REQUEST_ENTITIES, meta: { widgetId } });
-  let entityQueries = entityIds.map((id, index) => {
-    return `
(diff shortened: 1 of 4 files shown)


packages/veritone-widgets/src/redux/modules/mediaDetails/saga.js
```diff
@@ -637,40 +638,39 @@ function* fetchEntities(widgetId, entityIds) {
...
-      error: 'Error thrown while fetching entities',
-      meta: { widgetId }
+      error: 'Error fetching entities from server.'
     });
   }
   yield put({
     type: REQUEST_ENTITIES_SUCCESS,
-    payload: response.data,
-    meta: { widgetId }
+    payload: response.data.entities.records
```
ChrisPelletier: Should we use a get here? I don't think it will get here because of the return when there is an error in the response but just in case?

PR #282, opened 2018-11-01
VTN-11610 emit event when face entity is updated.

AlexOVeritone:
Link to Ticket: https://steel-ventures.atlassian.net/browse/VTN-11610

Reviewers
---------
@ChrisPelletier @atb91590 

Purpose
-------
For Wazee integration when user edits entity in the face engine output, or resets edit to original - CMS should notify them.

What Changed
------------
In MDP added saga that'll call GQL to emit EntityUpdated even.
Decided to put that in MDP code because 1. potentially entities could be updated in different components: logo, object, face, etc. 2. This is Veritone+Wazee specific usecase, so probable smart components should not emit such event if reused outside of MDP.

How should we monitor this?
---------------------------
MDP save Face/Transcript, restore original should work as it worked before. For event check nsq logs on the environments.
Keep in mind that event first has to be created manually as described in https://steel-ventures.atlassian.net/browse/VTN-11513

### DIFF
--- a/packages/veritone-widgets/src/redux/modules/mediaDetails/saga.js
+++ b/packages/veritone-widgets/src/redux/modules/mediaDetails/saga.js
@@ -23,10 +23,6 @@ import {
 } from 'lodash';
 import { helpers, modules } from 'veritone-redux-common';
 import {
-  pendingUserEdits,
-  ADD_DETECTED_FACE,
-  REMOVE_DETECTED_FACE,
-  CANCEL_FACE_EDITS,
   SAVE_FACE_EDITS_SUCCESS
 } from './faceEngineOutput';
 import {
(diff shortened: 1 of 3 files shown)


packages/veritone-widgets/src/redux/modules/mediaDetails/saga.js
```diff
@@ -886,6 +887,41 @@ function* watchRestoreOriginalEngineResults() {
...
+  const input = {
+    eventName: 'EntityUpdated',
+    eventType: 'entity',
+    application: '8a37c1d0-3f3b-48d0-a84e-2b8e3646fbe5',
+    payload: `{ "recordingId": "${tdoId}" }`
+  };
+
+  try {
+    if (tdoId) {
+      const response = yield call(fetchGraphQLApi, {
```
ChrisPelletier: Can we use callGraphQLApi here? https://github.com/veritone/veritone-sdk/blob/fdd6201b9e849b331c1bb436af0dffc60b2505f6/packages/veritone-widgets/src/redux/modules/engineOutputExport/index.js#L424

AlexOVeritone: Thank you. I'll address this in a separate commit.

PR #283, opened 2018-11-01
VTN-11800 Media player overlay auto select

SETA-thanhnt:


olslash:
this won't work because the IDs of bounding boxes in this component are actually supplied by the user of the component -- for example if I create a new bounding box and save it to the server, the server will assign an ID and i use _that id_ to identify the box.

so I'm not 100% sure how to handle auto-select without implementing it on the user's side

SETA-thanhnt:
@olslash 
Yes, currently bounding box ID is created by client app.
The idea here is creating bounding box ID from MediaPlayer, then return bounding box data (include ID) to client app. The app is able to override the ID if necessary.
I will modify Redact app logic after this PR is merged and released.

PR #294, opened 2018-11-14
VTN-11400 fail 24h+ incomplete tasks.

AlexOVeritone:
Link to Ticket: https://steel-ventures.atlassian.net/browse/VTN-11400

## Reviewers

@ChrisPelletier 

## Purpose
If task is not in one of the final statuses for 24h+ consider it failed. Do this on the client. Api will respect task status.

## Dependencies
GQL https://github.com/veritone/core-graphql-server/pull/1192

## Hygiene

* [ ] Tested in dev environment


AlexOVeritone:
yep

PR #298, opened 2018-12-03
VTN-9468 From a video media file after going to the "Trim and Process…

seta-huongdv:
https://steel-ventures.atlassian.net/browse/VTN-9468


JordanBelford:
@olslash this looks fine to me but let me know if you have any objections

PR #301, opened 2018-12-10
By Scene face detection filtering

ChrisPelletier:
https://steel-ventures.atlassian.net/projects/VTN/issues/VTN-11620

When we are displaying face detection and the users select By Scene in the engine output header dropdown we should be filtering both face detections and face recognitions.

### DIFF
--- a/packages/veritone-react-common/src/components/FaceEngineOutput/RecognizedFaceMatch/index.js
+++ b/packages/veritone-react-common/src/components/FaceEngineOutput/RecognizedFaceMatch/index.js
@@ -29,7 +29,7 @@ class RecognizedFaceMatch extends Component {
           <div className={styles.entityFullName}>{entity.fullName}</div>
         </div>
         <div>
-          {confidence && (
+          {!!confidence && (
             <div className={cx(styles.timeSlotConfidence, confidenceColor)}>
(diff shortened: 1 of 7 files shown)


packages/veritone-react-common/src/components/FaceEngineOutput/RecognizedFaceMatch/index.js
```diff
@@ -29,7 +29,7 @@ class RecognizedFaceMatch extends Component {
           <div className={styles.entityFullName}>{entity.fullName}</div>
         </div>
         <div>
-          {confidence && (
+          {!!confidence && (
```
dapengAtVeritone: why !! ?

ChrisPelletier: Because is confidence is undefined then react will render it as a 0. I could use lodashs !isUndefined(confidence) to make it more clear.

dapengAtVeritone: Thanks! I am a layperson to React so please do it in your way. :)  I saw `&&` behind it so thought confidence will be boolean for sure.

PR #309, opened 2019-01-23
Sdo integration changes to master

atb91590:
Focus day efforts to move all sdo-integration stuff into veritone-react-components.

There were some changes that made it to master at some point, but has since become unsynced. These components will be removed in favor of those in the new repo.

There were also some additional changes to components which will reside in master branch (eg. TimeRangePicker) that required additional changes (minor changes).


packages/veritone-react-common/src/components/TopBar/styles.scss
```diff
@@ -7,6 +7,7 @@
   justify-content: space-between;
   left: 0;
   right: 0;
+  margin-bottom: 20px;
```
olslash: this should be removed -- it doesnt make sense for a self contained component like this to impose a margin on the content below it. also it will mess up all other existing users of topbar.

packages/veritone-react-common/src/components/DataTable/MenuColumn.js
```diff
@@ -33,6 +34,7 @@ export default class MenuColumn extends React.Component {
     onSelectItem: noop,
     protectedActions: ['delete'],
     transformLabel: l => l,
+    transformActions: a => a,
```
olslash: seeing this error in the storybook:

image

packages/veritone-react-common/src/components/formComponents/DateTimePicker.js
```diff
@@ -102,22 +113,21 @@ TimeSelector.propTypes = {
...
+  return value ? (
+    <TextField
+      className={styles.dateTimeTZ}
+      value={value}
+      InputProps={{
+        disableUnderline: true
+      }}
+      disabled
+    />
+  ) : null;
```
olslash: pointless change, `value &&` is the correct pattern

packages/veritone-redux-common/src/helpers/api/fetchGraphQLApi.js
```diff
@@ -5,7 +5,9 @@ export default function fetchGraphQLApi({
   query,
   variables,
   operationName,
-  token
+  token,
+  dispatch,
+  getState
```
olslash: adds unused args

packages/veritone-widgets/src/build-entry.js
```diff
@@ -7,23 +7,20 @@ export OAuthLoginButton, {
   OAuthLoginButtonWidget
 } from './widgets/OAuthLoginButton';
 export FilePicker, { FilePickerWidget } from './widgets/FilePicker';
+export filePickerReducer, * as filePickerModule from './redux/modules/filePicker';
+export filePickerSaga from './redux/modules/filePicker/filePickerSaga';
 export { EngineSelectionWidget } from './widgets/EngineSelection';
-export { TableWidget } from './widgets/Table';
+export { TableWidget as Table, TableWidget  } from './widgets/Table';
```
olslash: Table would be a plain component, not a wrapped widget -- so this change doesnt make sense

packages/veritone-react-common/src/components/DataTable/MenuColumn.js
atb91590: made a recent commit to fix this issue (used omit)

packages/veritone-react-common/src/components/formComponents/DateTimePicker.js
atb91590: i recall the component actually needs a value returned (null as this object). otherwise errors will be thrown on undefined content.

packages/veritone-redux-common/src/helpers/api/fetchGraphQLApi.js
atb91590: ah, you right

packages/veritone-widgets/src/build-entry.js
atb91590: agreed. i'll remove it

packages/veritone-react-common/src/components/TopBar/styles.scss
atb91590: 👍

PR #315, opened 2019-02-25
Feature/vtn 14061 add previous and next button mdp

seta-tuha:
Ticket: https://steel-ventures.atlassian.net/browse/VTN-14061
What's change
- Add previous and next buttons
- Add onNext, onBack, isNextActive, isBackActive props


packages/veritone-widgets/src/widgets/MediaDetails/index.js
```diff
@@ -1067,9 +1085,42 @@ class MediaDetailsWidget extends React.Component {
                       {!isLoadingTdo && 'No Filename'}
                     </div>
                   )}
-                  <div 
+                    <IconButton
```
atb91590: hide this button if `onBack` is not defined

packages/veritone-widgets/src/widgets/MediaDetails/index.js
```diff
@@ -1067,9 +1085,42 @@ class MediaDetailsWidget extends React.Component {
...
+                          id="tooltip-back-media"
+                          title="Previous file"
+                      >
+                        <Icon
+                          classes={{ root: styles.iconClass }}
+                          className="icon-keyboard_arrow_left"
+                        />
+                      </Tooltip>
+                    </IconButton>
+                    <IconButton
```
atb91590: hide this button if `onNext` is not defined

atb91590:
changes communicated from jira ticket:
```when paginating across the 24 items, can we also keep CMS's pagination state in sync?

For example, MDP pages from item 24 -> 25 -> 26. CMS's page should be updated to have offset of 24 so that we are now on page 2 (viewing 24-47).

Also, the disabled button state issue is still the same. I recommend to disable the tooltip hover state and don't use cursor: pointer.```

PR #319, opened 2019-03-06
JSON schemas and validator function for object

atungw:


packages/veritone-json-schemas/schemas/vtn-standard/object/object.json
```diff
@@ -0,0 +1,82 @@
...
+                "type": "integer",
+                "minimum": 0
+              },
+              "stopTimeMs": {
+                "type": "integer",
+                "minimum": 0
+              },
+              "type": {
+                "type": "string",
+                "pattern": "object"
```
SteveShaffer: I think there's a `const` keyword you can use that's more explicit?

packages/veritone-json-schemas/schemas/vtn-standard/object/object.json
```diff
@@ -0,0 +1,82 @@
...
+              }
+            },
+            "required": [
+              "startTimeMs",
```
SteveShaffer: `type` should also be required for this, as should `object` (at least for this object validationContract)

packages/veritone-json-schemas/.babelrc
```diff
@@ -0,0 +1,6 @@
+{
```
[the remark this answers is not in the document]
atungw: Every package in veritone-sdk can use a different .babelrc, since it's probably too onerous to force all of it to use the same babel version. Frontend projects tend to me much more current than backend in terms of ES stage support.

packages/veritone-json-schemas/schemas/vtn-standard/master.json
```diff
@@ -0,0 +1,223 @@
...
+            "type": "string",
+            "description": "Private id variable used by the engine to refer to its internal dataset",
+            "example": "1"
+          }
+        },
+        "required": [
+          "class"
+        ]
+      },
+      "description": "Optional category for object recognition"
```
SteveShaffer: Optional categor**ies** for object recognition. Used for referencing external taxonomies.

packages/veritone-json-schemas/schemas/vtn-standard/master.json
```diff
@@ -0,0 +1,223 @@
...
+      "type": "object",
+      "properties": {
+        "boundingPoly": {
+          "$ref": "#/definitions/boundingPoly"
+        },
+        "confidence": {
+          "$ref": "#/definitions/confidence"
+        },
+        "entityId": {
+          "type": "number"
```
SteveShaffer: GUID.  So `"type": "string"`

packages/veritone-json-schemas/schemas/vtn-standard/master.json
```diff
@@ -0,0 +1,223 @@
...
+        },
+        "entityIdentifierId": {
+          "type": "number"
```
SteveShaffer: GUID.  So `"type": "string"`

packages/veritone-json-schemas/schemas/vtn-standard/master.json
```diff
@@ -0,0 +1,223 @@
...
+        },
+        "libraryId": {
```
SteveShaffer: We decided that we'd like to remove this from the spec since it's duplicate information with the already-unique `entityId`.  So the best engine developers can do by writing this value is not screw up.

packages/veritone-json-schemas/schemas/vtn-standard/master.json
```diff
@@ -0,0 +1,223 @@
...
+          "type": "number"
+        },
+        "label": {
+          "type": "string"
+        },
+        "type":
+          {
+            "type": "string",
+            "enum": [
```
SteveShaffer: This seems like the wrong list to me?

The list in the big example is:
```
      // - object: Object detection
      // - face: Face detection
      // - licensePlate: License plate detection
      // - logo: Logo detection
      // - fingerprint: Audio fingerprinting
      // - speakerId: Speaker recognition
      // - soundId: Sound recognition
      // - concept: Concept recognition
      // - keyword: Keyword detection
      // - text: Recognized or extracted text (OCR / text extraction)
      // - namedEntity: Entity extraction
      // - barcode
```

packages/veritone-json-schemas/schemas/vtn-standard/master.json
```diff
@@ -0,0 +1,223 @@
...
+              "audio-fingerprint",
```
SteveShaffer: This should at least be camelCase.  But actually I think for better or worse we're stuck with this as just `fingerprint`

packages/veritone-json-schemas/schemas/vtn-standard/master.json
```diff
@@ -0,0 +1,223 @@
...
+              "ocr",
```
SteveShaffer: `text`

packages/veritone-json-schemas/schemas/vtn-standard/master.json
```diff
@@ -0,0 +1,223 @@
...
+              "face",
+              "object",
+              "transcript"
```
SteveShaffer: transcripts are not objects.  Remove.

packages/veritone-json-schemas/schemas/vtn-standard/master.json
```diff
@@ -0,0 +1,223 @@
...
+        "y"
+      ],
+      "description": "A point describing the 2d coordinates from the top left corner of the source."
+    },
+    "boundingPoly": {
+      "type": "array",
+      "items": {
+        "$ref": "#/definitions/point"
+      },
+      "required": [
```
SteveShaffer: Is this correct?  This is already required on the `point` definition.

packages/veritone-json-schemas/schemas/vtn-standard/object/object.json
```diff
@@ -0,0 +1,37 @@
...
+      "items": {
+        "const": "object"
+      }
+    },
+    "object": {
+      "type": "array",
+      "items": {
+        "type": "object",
+        "properties": {
+          "object": {
```
SteveShaffer: Doesn't this mean something like this?
```
{
  "object": [{
    "object": {
      // objectResult definition goes here
    }
  }]
}
```

And if so, what's wrong with the tests that allows this to work?

packages/veritone-json-schemas/schemas/vtn-standard/object/object.json
```diff
@@ -0,0 +1,37 @@
...
+            "$ref": "master.json#/definitions/objectResult"
```
SteveShaffer: Are relative file references allowed?  And if so, shouldn't this be `"../master.json#/definitions/objectResult"`?

packages/veritone-json-schemas/schemas/vtn-standard/object/object.json
```diff
@@ -0,0 +1,37 @@
...
+        }
+      }
+    },
+    "series": {
+      "type": "array",
+      "items": {
+        "type": "object",
+        "properties": {
+          "object": {
+            "$ref": "master.json#/definitions/objectResult"
```
SteveShaffer: Same relative reference comments as above

packages/veritone-json-schemas/schemas/vtn-standard/transcript/transcript.json
```diff
@@ -0,0 +1,116 @@
...
+      "minItems": 1,
+      "if": {
+        "minItems": 2
+      },
+      "then": {
+        "items": {
+          "type": "object",
+          "properties": {
+            "word": {
+              "$ref": "master.json#/definitions/word"
```
SteveShaffer: Same relative reference comments as on object (for all `$ref`s)

packages/veritone-json-schemas/schemas/vtn-standard/transcript/transcript.json
```diff
@@ -0,0 +1,116 @@
...
+          ]
+        }
+      }
+    }
+  },
+  "properties": {
+    "validationContracts": {
+      "type": "array",
+      "items": {
+        "const": "transcript"
```
SteveShaffer: I don't think this allows for other `validationContracts` values to be present in the array (which should be allowed for it to still validate as a transcript).

packages/veritone-json-schemas/schemas/vtn-standard/transcript/transcript.json
```diff
@@ -0,0 +1,116 @@
...
+        "required": [
+          "startTimeMs",
+          "stopTimeMs"
+        ]
+      }
+    },
+    "language": {
+      "$ref": "master.json#/definitions/language"
+    },
+    "schemaId": {
```
SteveShaffer: Why does `transcript` define all these header fields but `object` doesn't?

packages/veritone-json-schemas/src/index.test.js
```diff
@@ -0,0 +1,65 @@
...
+import _ from 'lodash';
+
+import glob from 'glob';
+
+const categories = {};
+
+// dynamically import all json files in examples or invalid-examples
+const files = glob.sync("schemas/vtn-standard/**/**(examples|invalid-examples)/!(master.json)//*.json");
+for(let file of files) {
+  const categoryIndex = file.split("/").findIndex(directory => _.includes(directory, "examples")) - 1;
```
SteveShaffer: Naming convention nitpick: `category` should be `validationContract` or `contract` in the context of this vtn-standard spec.  "Category" is an aiWARE-specific notion.

packages/veritone-json-schemas/src/index.test.js
```diff
@@ -0,0 +1,65 @@
...
+
+Object.keys(categories).forEach(category => {
+  describe(`"${category}" tests`, () => {
+    const validator = VALIDATORS[category];
+    if(typeof validator === 'function') {
+      for(let test of categories[category]) {
+        if( _.includes(test.directory, 'invalid' )) {
+
+          // JSON files in invalid-examples should not validate
+          it(`should NOT validate ${test.directory}/${test.fileName}`, async () => {
```
SteveShaffer: I love that I can just add a new file in the proper folder/naming convention and it'll be picked up as a new test.  But the flip side of that of course is that the test errors are not very specific.  If there were any way to throw more specific error messages that'd be great.  If it's not easy, don't worry about it; the pros of having DRYer test case writing outweigh the cons of having to search for your errors IMO.

packages/veritone-json-schemas/schemas/vtn-standard/transcript/transcript.json
atungw: We pivoted from object, so I never got to finish it. We'll revisit it later with more tests and such.

packages/veritone-json-schemas/schemas/vtn-standard/transcript/transcript.json
atungw: This is on purpose for now, to ensure they are one to one and to make sure the correct value is provided.

My general plan for this is to create each sub schema one by one then merge them all into the master, otherwise if you develop on the master schema directly the matching for JSON schema makes it hard to accomplish, especially without breaking existing property matches.

packages/veritone-json-schemas/src/index.test.js
atungw: The alternative to this would be to make the examples in a .JS vs .JSON file, and then provide something like

const data = {
   ...VTNSAMPLE
}

const errors = {
   ERRORS_CAPTURED_FROM_LIBRARY
}

but I'm preferring .json because it makes it easier for people to understand and easier for people to import. Also, I'm not sure how stable the error reporting is from version to version of AJV, so the test might be brittle if I do that.

It's always possible to add your own specific test in the test file if you want to test a specific scenario (which I will probably do for additional properties)

packages/veritone-json-schemas/schemas/vtn-standard/master.json
```diff
@@ -0,0 +1,223 @@
...
+          "example": "fe261482-af8d-4709-aefb-ff1c19b61eb9"
+        },
+        "sourceEngineName": {
+          "type": "string",
+          "description": "Engine name used to generate output (provided by Veritone) // TO DO: Verify if this is internal or external engine name",
+          "example": "capio"
+        },
+        "taskPayload": {
+          "type": "object",
+          "description": "JSON key/value task payload describing the tasks used to start the engine run (provided by Veritone) // TO DO: LINK TO DOCS EXPLAINING TASK PAYLOAD",
```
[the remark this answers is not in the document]
SteveShaffer: Still pending TODO iteam

packages/veritone-json-schemas/schemas/vtn-standard/master.json
SteveShaffer: Please fix this before merging to master

packages/veritone-json-schemas/schemas/vtn-standard/master.json
SteveShaffer: Please fix this before merging to master

packages/veritone-json-schemas/schemas/vtn-standard/master.json
SteveShaffer: Please fix this before merging to master

packages/veritone-json-schemas/schemas/vtn-standard/master.json
SteveShaffer: Please fix this before merging to master

packages/veritone-json-schemas/schemas/vtn-standard/object/object.json
SteveShaffer: Can I get a response on this?  I'm particularly concerned on this one

packages/veritone-json-schemas/src/index.test.js
SteveShaffer: Tradeoffs.  Sounds good.

SteveShaffer:
prettify plz

packages/veritone-json-schemas/schemas/vtn-standard/master.json
```diff
@@ -0,0 +1,217 @@
...
+        },
+        "required": [
+          "class"
+        ]
+      },
+      "description": "Optional categories for object recognition. Used for referencing external taxonomies."
+    },
+    "objectResultType": {
+      "type": "string",
+      "enum": [
```
SteveShaffer: This is still wrong.  This is the validationContract enum.  The documented valid values for object.type are:
```
      // - object: Object detection
      // - face: Face detection
      // - licensePlate: License plate detection
      // - logo: Logo detection
      // - fingerprint: Audio fingerprinting
      // - speakerId: Speaker recognition
      // - soundId: Sound recognition
      // - concept: Concept recognition
      // - keyword: Keyword detection
      // - text: Recognized or extracted text (OCR / text extraction)
      // - namedEntity: Entity extraction
      // - barcode
```

packages/veritone-json-schemas/schemas/vtn-standard/master.json
```diff
@@ -117,11 +117,18 @@
...
-        "audio-fingerprint",
-        "ocr",
-        "face",
         "object",
-        "transcript"
+        "face",
+        "licensePlate",
+        "logo",
+        "fingerprint",
+        "speakerId",
```
SteveShaffer: change to `speaker`

packages/veritone-json-schemas/schemas/vtn-standard/master.json
```diff
@@ -117,11 +117,18 @@
...
+        "soundId",
```
SteveShaffer: change to `sound`

PR #329, opened 2019-04-05
[VTN-20155] expanding libraries index in searchbar

seta-tuha:


packages/veritone-react-common/src/components/SearchBar/component.js
```diff
@@ -365,33 +368,49 @@ export class SampleSearchBar extends React.Component {
...
-          `query {
-            libraries(limit: 50) {
-              records {
-                id
-                name
-                description
-                coverImageUrl
+      let LIBRARY_OFFSET = 0;
+      let libraries = [];
+      while (true) {
```
atb91590: Using this type of loop presents a HUGE risk of infinitely looping if the service fails to respond correctly in any case. Instead, you should abstract the get library call into a separate function which returns a promise and chain the promise returns until we reach the library limit.

PR #333, opened 2019-04-20
VTN-21354:move-scroll-inside-the-image-list

seta-linhle:
Link to ticket https://steel-ventures.atlassian.net/browse/VTN-21354
Link to ticket https://steel-ventures.atlassian.net/browse/VTN-21353
Reviewer
@atb91590

Purpose
Fix bug: VTN-21354 and VTN-21353

### DIFF
--- a/packages/veritone-react-common/src/components/FilePicker/FileList/styles.scss
+++ b/packages/veritone-react-common/src/components/FilePicker/FileList/styles.scss
@@ -15,6 +15,8 @@ $preview-size: 40px;
   position: relative;
   overflow-y: scroll;
   margin-left: 15px;
+  overflow: auto;
+  height: 100%;
 }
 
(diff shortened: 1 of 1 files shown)


packages/veritone-react-common/src/components/FilePicker/FileList/styles.scss
```diff
@@ -15,6 +15,8 @@ $preview-size: 40px;
   position: relative;
   overflow-y: scroll;
   margin-left: 15px;
+  overflow: auto;
```
atb91590: just update the overflow-y above to `auto` instead of overriding it here

seta-linhle: updated.

PR #342, opened 2019-05-08
Feature/VTN-20276 Multiple engine picker and selection info panel

atungw:


packages/veritone-widgets/src/redux/modules/multipleEngineSelection/index.js
```diff
@@ -0,0 +1,223 @@
...
+        },
+        count
+      },
+      me {
+        organizationId
+      }
+    }
+  `;
+
+  return await callGraphQLApi({
```
gitrite: not a real issue but you can omit `await` if you're returning from an asynchronous operation inside an async function

packages/veritone-widgets/src/widgets/EnginePicker/MultiplePicker.js
```diff
@@ -0,0 +1,126 @@
...
+  static propTypes = {
+    canSelectMore: bool,
+    selected: objectOf(any),
+    toggleSelection: func
+  };
+
+  constructor(props) {
+    super(props);
+
+    this.id = guid();
```
gitrite: i'm thinking more descriptive property names would help here? just to make the code more self-documenting

packages/veritone-widgets/src/widgets/EnginePicker/SelectionInfoPanel.js
```diff
@@ -0,0 +1,121 @@
...
+    return (
+      <Paper elevation={4} style={{ height: '100%', minHeight: '300px' }}>
+        <div className={styles.pickerTitle}>
+          <Typography variant="headline">
+            Selected Engines{' '}
+            <Typography
+              variant="subheading"
+              style={{ display: 'inline-block' }}
+            >
+              {Object.keys(this.props.selected).length} / 6
```
gitrite: hard coded values scare me. can we put it in a variable that helps to make it clear what it's for? additionally, is there value in making it a prop?

atungw: Oops, yeah I meant to make that a prop.

PR #343, opened 2019-05-09
Create InfinteDirectoryList scroll

seta-tuha:


atb91590:
@seta-tuha Please add a storybook

atungw:
This should also not be using setTimeout, don't use setState inside of a setTimeout.

packages/veritone-react-common/src/components/InfiniteDirectoryList/index.js
```diff
@@ -0,0 +1,182 @@
...
+                    <div className={styles['table-row-text']}>
+                      {header}
+                    </div>
+                  </TableCell>
+                ))
+              }
+            </TableRow>
+          </TableHead>
+          <TableBody className={classes.tablebody}>
+            {files.map(({ id, type, name, date, selected }) => {
```
atb91590: I've updated the specs to use id/type. `type` will either be 'folder' or 'tdo'. When it is 'tdo, you will have to look into `file.primaryAsset.contentType` to figure out the which icon to use.

Here is an example of how you can determine which icon to use: 
https://github.com/veritone/veritone-sdk/blob/master/packages/veritone-react-common/src/components/FilePicker/FileProgressList/index.js#L42

packages/veritone-react-common/src/components/InfiniteDirectoryList/story.js
```diff
@@ -0,0 +1,41 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import { action } from '@storybook/addon-actions';
+import InfiniteDirectoryList from './';
+
+const items = [
+  { id: '1', type: 'Folder', date: 'Mar 29, 2019 3:34 PM', name: 'A folder' },
```
atb91590: update your story to conform to the correct format of items in the spec

packages/veritone-react-common/src/components/InfiniteWrapper/styles.scss
```diff
@@ -0,0 +1,4 @@
+.infinte-wrapper {
+  height: 300px;
```
atb91590: Do not use a static height. Height should be inherited from its parent and still work with the scroll loader

packages/veritone-react-common/src/components/InfiniteDirectoryList/styles.scss
```diff
@@ -0,0 +1,20 @@
+.loading-container {
```
atb91590: Use camel case class names. This allows easier access to styles (styles.loadingContainer).

`loading-container` => `loadingContainer`

seta-tuha: Actually, scss convention className is not using camelCase
http://bdavidxyz.com/blog/how-to-name-css-classes/

seta-tuha:
Merged into Data Picker

PR #344, opened 2019-05-10
Add Leftnavigationpanel

seta-tuha:


packages/veritone-react-common/src/components/LeftNavigationPanel/index.js
```diff
@@ -0,0 +1,79 @@
...
+          </ListItem>
+        </React.Fragment>
+      ))
+    }
+  </List>
+);
+
+
+LeftNavigationPanel.propTypes = {
+  pathList: arrayOf(shape({
```
atb91590: Are you not referring to the tech specs at all when building these components?
https://steel-ventures.atlassian.net/wiki/spaces/VT/pages/949289104/Data+Picker+Tech+Specs#DataPickerTechSpecs-Task3.5:LeftNavigationpanel

seta-tuha: @atb91590 The problem is, we should make the component be general for usage. In case we parse the props like in your specs, only cms can use this component

atb91590: There are no other components planned to use this component. Please follow the tech specs

PR #347, opened 2019-05-13
Add media info component

seta-tuha:
[WIP] Using CSSTransitionGroup


packages/veritone-react-common/src/components/MediaInfoPanel/SimpleMediaPlayer.js
```diff
@@ -0,0 +1,84 @@
+import React from 'react';
+import { arrayOf, string, shape } from 'prop-types';
+
+import shaka from 'shaka-player';
```
atb91590: You should not be reimplementing your own media player. Please use the existing media player found in veritone-sdk. However, the that media player in `veritone-sdk` master branch needs to be updated/replaced with the version in `veritone-react-components`.

In addition we need to demote the `MediaPlayer` to veritone-react-common so that it can be used within your media info panel

atb91590: I'm currently working on migrating the video player to react-common so that you can use it

seta-tuha: @atb91590 I see, the video-player will need to improve. I will contribute to your PR

atb91590: @seta-tuha I've just about finished the media player migration from veritone-react-components => veritone-sdk. Here is the PR:
https://github.com/veritone/veritone-sdk/pull/349

Once these changes have been merged to master, you should be able to replace your implementation of the media player (look at storybook for how to use it)

atb91590:
I just noticed this component was not scoped for the MVP .... please do not prioritize this component

PR #351, opened 2019-05-15
Feature/vtn 22864 01 create upload button

seta-tuha:


atb91590:
@seta-tuha You do not need to create an entirely new component just for a single button... This button should just live IN the header component

PR #352, opened 2019-05-15
Feature/vtn 22864 02 create search input component

seta-tuha:


packages/veritone-react-common/src/components/SearchInput/index.js
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import { func, string } from 'prop-types';
+import SearchIcon from '@material-ui/icons/Search';
+import ClearIcon from '@material-ui/icons/Clear';
+
+import styles from './styles.scss';
+
+const SearchInput = ({ onChange, onClear, value }) => (
```
atb91590: You should wrap the onChange function with a handler that debounces the users input so that searches don't get triggered on every keystroke

MmtBkn:
Closing as the primary ticket is closed.

PR #353, opened 2019-05-15
Add back button component

seta-tuha:


atb91590:
Do not create separate files for such simple components. Just put this in the header component

PR #354, opened 2019-05-15
[WIP] Feature/vtn 22864 create header bar

seta-tuha:
This pr consolidates from other prs


packages/veritone-react-common/src/components/BackButton/index.js
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
```
atb91590: Put this directly into the header, you don't need to separate it. It's a simple button...

packages/veritone-react-common/src/components/HeaderBar/index.js
```diff
@@ -0,0 +1,216 @@
...
+              ))
+            }
+          </MenuList>
+        </Popover>
+      </div>
+    );
+  }
+}
+
+HeaderBar.propTypes = {
```
atb91590: move the propTypes into the HeaderBar class body

packages/veritone-react-common/src/components/HeaderBar/index.js
```diff
@@ -0,0 +1,216 @@
...
+    id: string
+  })),
+  onCrumbClick: func,
+  onSearch: func,
+  onClear: func,
+  onBack: func,
+  isStream: bool,
+  viewType: string,
+  onToggleView: func,
+  onUpload: func,
```
atb91590: clicking the upload button uses the function onToggleView (aka `toggleViewType` as in the specs) the change the view to the upload screen like so:

```toggleViewType('upload')```

File: packages/veritone-react-common/src/components/HeaderBar/index.js [hunk 4]
```diff
@@ -0,0 +1,216 @@
...
+  }), {})),
+  pathList: arrayOf(shape({
+    label: string,
+    icon: string,
+    id: string
+  })),
+  onCrumbClick: func,
+  onSearch: func,
+  onClear: func,
+  onBack: func,
```
atb91590: Since we're not planning to retain a history of the previous views... lets just make the `onBack` be an internal function (not a prop) that will execute `toggleViewType('folder')`

packages/veritone-react-common/src/components/BackButton/index.js
seta-tuha: done

packages/veritone-react-common/src/components/HeaderBar/index.js
seta-tuha: done

PR #356, opened 2019-05-16
Udate video source as react api change

seta-tuha:


MmtBkn:
@seta-tuha Is there a ticker for this PR? If It's not going to merge, please close it.

PR #376, opened 2019-07-08
feature/vtn-26250 Feature: Create Reusable Simple Search Bar

MmtBkn:
https://steel-ventures.atlassian.net/browse/VTN-26250


packages/veritone-react-common/src/components/SimpleSearchBar/index.js
```diff
@@ -0,0 +1,221 @@
...
+              <div style={{ overflow: 'scroll' }}>
+                {autocompleteResults.map(result => {
+                  // if we have plain string array
+                  // eslint-disable-next-line lodash/prefer-lodash-typecheck
+                  if (typeof result === 'string') {
+                    return (
+                      <MenuItem
+                        key={result}
+                        // eslint-disable-next-line react/jsx-no-bind
+                        onClick={() => clear() || result.onClick(result)}
```
twoo24: Won't `onClick` here result in an error if `result` is `string`?

packages/veritone-react-common/src/components/SimpleSearchBar/index.js
```diff
@@ -0,0 +1,221 @@
...
+                                    clear() || result.onClick(data)
+                                  }
+                                >
+                                  {TemplateComponent(data)}
+                                </MenuItem>
+                              );
+                            } else {
+                              return (
+                                <MenuItem
+                                  key={data.id}
```
twoo24: Is the `id` always going to be available on `data`?  Curious if there was an error when rendering in storybook since the data objects in `entities.json` don't have `id`s

packages/veritone-react-common/src/components/SimpleSearchBar/index.js
MmtBkn: Result is actually string here. If onClick is null, that would throw error, but we wouldn't care much, would we?

packages/veritone-react-common/src/components/SimpleSearchBar/index.js
MmtBkn: Right, data should have id. Otherwise react will throw every element should have unique key, but it will assign one, and won't cause any problem. I was going to make it required, but since different apps will be passing different data and templates here, that's not possible with prop types.

MmtBkn:
@twoo24 added new commits.

packages/veritone-react-common/src/components/SimpleSearchBar/index.js
```diff
@@ -0,0 +1,227 @@
...
+
+  state = {
+    renderKey: 101,
+    isLoading: true,
+    focused: false
+  };
+
+  clickAway = () => {
+    if (this.props.resetOnClickAway) {
+      this.props.onChange('');
```
SteveShaffer: Lines 36-40 should just be `this.clear()` to make sure all reset logic is done consistently?

packages/veritone-react-common/src/components/SimpleSearchBar/index.js
```diff
@@ -0,0 +1,227 @@
...
+}
+
+SimpleSearchBarBase.propTypes = {
+  autocomplete: bool,
+  onChange: func,
+  onSubmit: func,
+  value: string,
+  isLoading: bool,
+  placeholder: string,
+  autocompleteResults: arrayOf(any),
```
SteveShaffer: Should spec out the expected shape?

MmtBkn: Correct, resolved here; https://github.com/veritone/veritone-sdk/pull/382/commits/2c4fa7c07fff7dee90a89b5cbc285494f61dd409

PR #397, opened 2019-08-12
Feature/VTN-27517-Update appbar notifications

seta-thinhtd:
Link to ticket: https://steel-ventures.atlassian.net/browse/VTN-27517

Add feature to Notifications in veritone-sdk to have ability to:

   - Take count number as input and display as a the notif count (not use notifications.length as default anymore)
   - Can customize dialog from app


seta-linhle:
@seta-thinhtd  please update code to fix lint.

PR #399, opened 2019-08-20
VTN-26951 Add basic support for additional vtn-standard json-schemas

SteveShaffer:
Includes some relevant refactoring and an attempt at triggering CI
integration with now.


ctsiokos-veritone:
Since [VTN-26951](https://steel-ventures.atlassian.net/browse/VTN-26951) is somewhat broad I thought it might be ok to provide comments here on the overall vtn-standard:

vtn-standard/transcript/transcript.json
- @xyxxy13, note that when two or more words are provided the "bestPath" boolean field is currently a requirement. Do you foresee any issues with any transcription engines not providing a best path?
- When a single word is required, shouldn't the requirements for the presence of the "confidence" and "utteranceLength" field remain the same? When splicing is enabled, the OW can actually use the confidence field even with a single word present. That's because it can compare a pair of words coming from overlapping segments

vtn-standard/object/object.json
- We'd lose backwards compatibility but just pointing out that following the prevalent naming convention of arrays in the standard, the two "object" arrays, one at the top level and one in "series, should be named "objects"
- Following the convention in transcript.json to not require the "words" field in series items, as it may be empty, wouldn't it be better to also not require the "object(s)" field for the same reason? Vise-versa, if we require the "object(s)" field wouldn't it be better to require the "words" field?
- 

vtn-standard/master.json
- Compared to an older version ~9 months ago, the "tags" field is no longer present. I remember that @quynhdang-vt was using it. Quynh, any comments on this change?

xyxxy13:
@ctsiokos-veritone I don't know if every transcription engine can provide "bestPath".  It is certainly known in the engine internally but it may not always be part of its output.  And it can't be deduced from the confidences because the bestPath may not include the highest confidence word at a given time step.

xyxxy13:
What is utteranceLength used for?  If I have a word like "football" from time 0 to 2 along with "foot" 0-1 and "ball" 1-2, I assume utteranceLength should be 2 since there are other words that intersect that time span of "football".  But what downstream processing needs that and can't use timestamps instead?

SteveShaffer:
I have finished my work on this and it is now ready for review.

I have heavily refactored this to validate data very tightly with an increased disdain for additional properties outside the json-schema or properties missing from it.  Consumers who validate engine output are welcome to consider those cases valid within their context if they'd like (e.g. engine output of `{}` can be perfectly acceptable within EDGE).  But from the context of the json-schemas, everything must have `validationContracts` filled out and it must conform to that validation contract exclusively.

This helps us compose these schemas and paves the way for us to someday relax these constraints and allow for things like multi-validationContract outputs.  But at this time, we're keeping it tight.

This brings the full list of supported validationContracts to:
- `concept`: Content Classification
- `entity`: Entity Extraction
- `keyword`: Keyword Extraction
- `language`: Language Identification
- `media-translated`: Document (.txt, .docx, etc.) Translation
- `object`: Object Detection and Recognition
- `sentiment`: Textual Sentiment Analysis
- `summary`: Textual Summarization
- `text`: Text Extraction, OCR, Extracted Text Translation, and OCR Translation
- `transcript`: Transcription and Transcript Translation

That means the remaining validationContracts that are documented in docs.veritone.com but not defined here are:
- `audio-fingerprint`: Audio Fingerprinting
- `face`: Face Detection and Recognition
- `structured-data`: Data Correlation
- `geolocation`: Geolocation
- `speaker`: Speaker Detection and Recognition
- `license-plate`: License Plate Detection and Recognition
- `logo`: Logo Detection and Recognition

Most of these (with the possible exceptions of structured-data and geolocation) should be relatively straightforward to add.

SteveShaffer:
If there are no further comments on this by EOD tomorrow I'm going to consider this "good enough," merge it in, and publish it as a new version of the veritone-json-schemas npm package (either 0.2.0 or 1.0.0).

PR #401, opened 2019-09-03
Migrate react to version 16.9.0 - with hooks / material-ui to v4.4

seta-tuha:
Migrate react version and material-ui version


seta-tuha:
> @seta-tuha I'm pretty sure that there are so many breaking changes between material ui v1 and v4, and this will require larger code refactoring.
> 
> Please address breaking changes.

Hi @MmtBkn Actually the build is failed but we can run the story. All the code in this PR is completed

twoo24:
@seta-tuha @SteveShaffer @KhuongVeritone I want to approve this so badly, but I think we need to regression test a bit more (probably as a group?).  My understanding is that the breaking changes from 1.x to 3.x were minimal, but there are quite a few going from 3.x to 4.x.  I ran storybook on `veritone-react-common` and there were a couple of errors that showed in the console immediately regarding  `jss.options.createGenerateClassName` and `JssProvider`.  I also briefly looked through this [migration guide](https://material-ui.com/guides/migration-v3/) and some changes are not apparent just by running storybook, but if you go through each story, you'll start noticing some errors showing on the console.  For example, the `SvgIcon`s are now using the `htmlColor` prop to pass in the color of the icon; the icons we have are using still use the `nativeColor` prop and will not display the correct color.  Same goes for the `Button` component with new `variant` values, the `Grid` component with new `spacing` values, and others.

seta-tuha:
Thanks @MmtBkn and @twoo24 I have updated the components follow https://material-ui.com/guides/migration-v3/

PR #402, opened 2019-09-04
Form builder boilerplate

seta-tuha:
Create form builder template

Check readme: https://github.com/veritone/veritone-sdk/tree/form-builder-boilerplate/packages/veritone-react-common/src/components/FormBuilder


seta-tuha:
Move to forks as it will has large change

PR #404, opened 2019-09-09
Feature/vtn 28641 migrate react component

seta-tuha:
Ticket: https://steel-ventures.atlassian.net/browse/VTN-28641
Purpose: Migrate react to version 16.9.0 and material-ui to 4.4

PR #405, opened 2019-09-09
Feature/vtn 29398 create form blocks

seta-tuha:
Ticket https://steel-ventures.atlassian.net/browse/VTN-29398
Depend on: https://github.com/veritone/veritone-sdk/pull/404


packages/veritone-react-common/src/components/FormBuilder/FormBlocks/styles.scss
```diff
@@ -0,0 +1,33 @@
+.block-container {
```
MmtBkn: Format code

PR #411, opened 2019-09-13
Update after reviewing and add editer permission

seta-linhle:
Link to ticket:
https://steel-ventures.atlassian.net/browse/VTN-29346
https://steel-ventures.atlassian.net/browse/VTN-29583

Purpose: 
update the searchbar follow the ticket

### DIFF
--- a/packages/veritone-react-common/src/components/SearchBar/SearchBarContainer.js
+++ b/packages/veritone-react-common/src/components/SearchBar/SearchBarContainer.js
@@ -1,7 +1,7 @@
 import React from 'react';
 import { arrayOf, func, object, string, oneOf } from 'prop-types';
 import cx from 'classnames';
-import { get, uniq, isEmpty, isEqual, sortBy } from 'lodash';
+import { get, uniq, isEmpty, isEqual, sortBy, includes } from 'lodash';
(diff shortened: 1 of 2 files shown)


packages/veritone-react-common/src/components/SearchBar/SearchBarContainer.js
```diff
@@ -490,7 +508,7 @@ class SearchBarContainer extends React.Component {
         ...customMenuActions,
       ]
     } else {
-      if (this.props.searchParameters && this.props.searchParameters.length > 0) {
+      if (this.props.searchParameters && this.props.isEditor && this.props.searchParameters.length > 0) {
         customMenuActions = [
           { label: 'Load Search Profile', onClick: (e) => { this.props.showLoadSavedSearch(e); this.handleMenuClose() } },
```
MmtBkn: They should be able to load profile, but not save.

seta-linhle: It will switch to else block. So I think it's will be same

MmtBkn: I see.

PR #414, opened 2019-09-18
Feature/vtn 29400 create form builder utils

seta-tuha:


packages/veritone-react-common/src/components/FormBuilder/Form.js
```diff
@@ -0,0 +1,42 @@
...
+        })
+      }
+    </div>
+  )
+}
+
+Form.propTypes = {
+  formDefinition: arrayOf(shape({ name: string })),
+  onChange: func,
+  value: objectOf(any),
```
MmtBkn: Could we be more specific?

packages/veritone-react-common/src/components/FormBuilder/FormItems/CheckBoxes.js
```diff
@@ -32,14 +32,20 @@ export default function Checkboxes({
   }, [name, value, onChange])
 
   return (
-    <FormControl error={error.length > 0} fullWidth>
+    <FormControl
+      error={error.length > 0}
```
MmtBkn: Could error be undefined here?

seta-tuha: all error is string and have default value to ''

PR #415, opened 2019-09-18
VTN-29336 [veritone-sdk] Add method to update widget props

MmtBkn:


packages/veritone-widgets/src/shared/PropUpdater.js
```diff
@@ -0,0 +1,24 @@
...
+import { func, object } from 'prop-types';
+
+export class PropUpdater extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {};
+    this.state.props = props.initialProps || {};
+  }
+
+  updateProps = newProps => {
```
twoo24: Is this getting used anywhere?

MmtBkn: That is the actual code :)

It's called from outside of component.

PR #419, opened 2019-09-24
Update changelog and increment ver for PR#403

mikomonday:
changes made were not documented and react-common was not incremented to 8.0.4


MmtBkn:
Hi @mikomonday, there were potential breaking changes(like material ui / react version updates). 

Can we revert this, un publish it (if we did), and do major release?

SteveShaffer:
yes please follow semver rules.  I took off the "require review from codeowners" rule so that i wouldn't be a bottleneck.  Don't make me turn it back on..... 😠

PR #420, opened 2019-09-26
Revert pull request #419 and remove v8.0.4 for react-common

mikomonday:
Included breaking issues with mui/react upgrades, reverting.

### DIFF
--- a/packages/veritone-react-common/CHANGELOG.md
+++ b/packages/veritone-react-common/CHANGELOG.md
@@ -228,8 +228,5 @@
 ## 8.0.2
 * Fixed AppBar css overridden issue
 
-## 8.0.3
+## 8.0.3 
 * Add Dosis font to global style import for the AppBar title
-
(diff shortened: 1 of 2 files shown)


mikomonday:
also unpublished veritone-react-common@8.0.4

SteveShaffer:
I wasn't aware npm would LET you unpublish a package (post-leftpad...).  I would have assumed you'd need to publish a new 8.0.5 reverting the change.  In any case, if they did, that's cool.  LGTM

mikomonday:
> I wasn't aware npm would LET you unpublish a package (post-leftpad...). I would have assumed you'd need to publish a new 8.0.5 reverting the change. In any case, if they did, that's cool. LGTM

Yea there was some text saying anyone using the package would be affected (ie leftpad) but since me and Adam were the only ones using it we saw no ill effects in doing so lol.

PR #429, opened 2019-10-17
VTN-30975 - CRUD folder dialog

seta-linhle:
Link to ticket:
https://steel-ventures.atlassian.net/browse/VTN-30975


packages/veritone-react-common/src/components/FolderTree/Modals/EditFolder/index.js
```diff
@@ -0,0 +1,115 @@
...
+import { get, isNil } from 'lodash';
+import { shape, bool, func } from 'prop-types';
+import TextField from '@material-ui/core/TextField';
+import Dialog from '@material-ui/core/Dialog';
+import DialogActions from '@material-ui/core/DialogActions';
+import DialogContent from '@material-ui/core/DialogContent';
+import DialogTitle from '@material-ui/core/DialogTitle';
+import styles from './styles.scss';
+
+export default function ModifyFolder({
```
twoo24: Can you clarify in which scenario a user will open the `ModifyFolder` modal vs. the `EditFolder`modal.  Seems like there's some duplicate logic there, but if it's different enough that should be fine.  This component should also be rename to `EditFolder` right?

packages/veritone-react-common/src/components/FolderTree/Modals/index.js
```diff
@@ -0,0 +1,9 @@
+import DeleteFolder from './DeleteFolder';
+import ModifyFolder from './DeleteFolder';
+import CreateFolder from './CreateFolder';
+
+export {
+    DeleteFolder,
+    ModifyFolder,
+    CreateFolder
```
twoo24: Is there a reason why EditFolder is not include here?

seta-linhle: It's a temp file I created and forgot deleting.

packages/veritone-react-common/src/components/FolderTree/Modals/EditFolder/index.js
seta-linhle: My first idea is breaking the CRUD into 4 components: Create, Edit, Move, Delete and they are not related to each other. Then after integrating into Collection and Folder Tree must be supported to multiple select folders, I made the Move Folder Modal support to create a new folder with selecting the parent folder. That's leading to the problem. I've removed the old edit modal and change the modify modal's name. thank you for reviewing.

PR #451, opened 2019-12-26
VTN-32312: update material ui latest

seta-tuha:
convert scss to jss


packages/veritone-widgets/src/styles/modules/variables.js
```diff
@@ -0,0 +1,50 @@
+// colors
```
twoo24: Where are these styles coming from?

seta-tuha: This is global styles.

seta-tuha:
@twoo24 I have merged some sub-branches. Can you re-review it?

PR #459, opened 2020-01-03
VTN-32314: upgrade materialUI, react, redux, storybook, babel

seta-tuha:


packages/veritone-react-common/src/styles/modules/muiTypography.js
```diff
@@ -0,0 +1,79 @@
+// https://github.com/muicss/mui/blob/master/src/sass/mui/mixins/_typography.scss
+
+export const muiTextStyles = {
```
MmtBkn: Set this props to theme

packages/veritone-react-common/.babelrc
```diff
@@ -1,7 +1,42 @@
 {
+  "presets": [
+    "@babel/preset-env",
```
MmtBkn: Add @babel/preset-typescript

MmtBkn:
@seta-tuha this task definition doesn't include upgrading redux / babel. Please be careful with the scope of the task next time.

seta-tuha:
> @seta-tuha this task definition doesn't include upgrading redux / babel. Please be careful with the scope of the task next time.

I got it. I will be paying more attention

packages/veritone-react-common/src/styles/modules/muiTypography.js
seta-tuha: work like this?
[https://material-ui.com/customization/theming/#custom-variables](url)

PR #483, opened 2020-03-30
Feature/es 358

seta-linhle:
Link to ticket:
https://steel-ventures.atlassian.net/browse/ES-358

### DIFF
--- a/packages/veritone-widgets/src/widgets/FilePicker/index.js
+++ b/packages/veritone-widgets/src/widgets/FilePicker/index.js
@@ -74,7 +74,9 @@ class FilePicker extends React.Component {
     onPickCancelled: func,
     onPick: func,
     height: number,
-    width: number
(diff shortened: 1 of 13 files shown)


packages/veritone-widgets/src/widgets/FilePicker/index.js
```diff
@@ -83,7 +85,9 @@ class FilePicker extends React.Component {
     onPick: noop,
     percentByFiles: [],
     height: 450,
-    width: 600
+    width: 600,
+    enableResize: false,
+    aspectRatio: 16/9,
```
MmtBkn: Would square be a better default here? It's mostly for avatars right?

seta-linhle: yeah. I will change it to 1/1.

PR #491, opened 2020-04-17
Feature/es 498

seta-linhle:
Link to ticket:
https://steel-ventures.atlassian.net/browse/ES-498

### DIFF
--- a/packages/veritone-widgets/src/widgets/FolderTree/index.js
+++ b/packages/veritone-widgets/src/widgets/FolderTree/index.js
@@ -56,6 +56,7 @@ function FolderTreeWrapper({
   initFolder,
   expandFolder,
   foldersData,
+  searchValue,
   originFolderData,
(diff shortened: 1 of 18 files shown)


packages/veritone-widgets/src/widgets/FolderTree/index.js
```diff
@@ -245,19 +254,21 @@ function FolderTreeWrapper({
...
     }
-    handleCloseModify();
   };
 
   const handleCloseModify = () => {
     setOpenModify(false);
-    setOpenEdit(false);
+    setTimeout(() => {
+      setOpenEdit(false);
+    }, 500)
```
MmtBkn, line 271: Why do we have a delay here?

seta-linhle: I want the dialog is disappeared completedly before we change the data.

PR #503, opened 2020-07-23
ES-677 Fix Year is not displayed in account last updated password

seta-tuha:
Link to ticket: https://steel-ventures.atlassian.net/browse/ES-677

### DIFF
--- a/packages/veritone-widgets/package.json
+++ b/packages/veritone-widgets/package.json
@@ -25,7 +25,8 @@
     "redux-thunk": "^2.3.0",
     "shaka-player": "^2.4.1",
     "veritone-oauth-helpers": "^2.1.0",
-    "video-react": "~0.13.3"
+    "video-react": "~0.13.3",
+    "reselect": "^4.0.0"
   },
   "main": "dist/bundle-cjs.js",
   "module": "dist/bundle-es.js",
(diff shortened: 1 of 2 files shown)


packages/veritone-widgets/package.json
```diff
@@ -25,7 +25,8 @@
     "redux-thunk": "^2.3.0",
     "shaka-player": "^2.4.1",
     "veritone-oauth-helpers": "^2.1.0",
-    "video-react": "~0.13.3"
+    "video-react": "~0.13.3",
+    "reselect": "^4.0.0"
```
MmtBkn, line 29: Why did we add reselect, remove it if not used.

packages/veritone-widgets/src/widgets/UserProfile/PasswordField/index.js
```diff
@@ -15,7 +15,7 @@ const PasswordField = ({ lastUpdated, onEdit }) => {
     <div className={classes.container}>
       <RaisedTextField
         label="Password"
-        value={`Last updated ${format(new Date(lastUpdated), 'MMMM dd, yyyy')}`}
+        value={`Last updated ${format(new Date(lastUpdated), 'MMMM DD, YYYY')}`}
```
MmtBkn, line 18: Problem is not the format, this value is being null. Please check app repos to see why wwe are getting this value null. We also like to show different message if password is never been updated. Something like;  

```suggestion
        value={lastUpdated ? `Last updated ${format(new Date(lastUpdated), 'MMMM DD, YYYY')}` : 'Never Updated'}
```

packages/veritone-widgets/package.json
seta-tuha: I saw it at https://github.com/veritone/veritone-sdk/blob/master/packages/veritone-widgets/src/redux/modules/folder/selector.js#L1

packages/veritone-widgets/src/widgets/UserProfile/PasswordField/index.js
seta-tuha: I logged value of `lastUpdated`, it's not null.

MmtBkn:
@seta-tuha is this released?

PR #508, opened 2020-08-07
ES-871 [Collections] Folder/Collection name can contain ONLY space

seta-tuha:
Link to ticket [ES-871](https://steel-ventures.atlassian.net/browse/ES-871)


MmtBkn:
@kenakaiser can you include this with the new release?

Thanks.

PR #527, opened 2020-10-12
feature/es-1096 [Library] [Local-feature-ES-807] Improper text color …

nathanng-veritone:
…Edit Profile button
https://steel-ventures.atlassian.net/browse/ES-1096
* update text color of EditProfileButton to white
* update CHANGELOG
* update veritone-react-common version to 10.0.4


MmtBkn (approved):
LGTM

PR #577, opened 2023-01-30
Feature/GLC-3817-Illuminate - Indiana State Police transcript timing out after 3 hours

seta-namle:
Link to ticket: https://steel-ventures.atlassian.net/browse/GLC-3817

### DIFF
--- a/packages/veritone-react-common/src/components/MediaPlayer/VideoSource.js
+++ b/packages/veritone-react-common/src/components/MediaPlayer/VideoSource.js
@@ -44,9 +44,7 @@ export default class VideoSource extends React.Component {
       video.load();
     }
     if (this.state.isStreamUriExpired !== prevState.isStreamUriExpired) {
-      this.setState({
-        streamUri: undefined,
-      });
+      this.setStreamUri(undefined);
       this.loadVideo(video, src, streams, disablePreload, currentTime);
     }
   }
@@ -102,9 +100,7 @@ export default class VideoSource extends React.Component {
           console.log('error loading video with shaka player', err);
         });
       }
-      this.setState({
-        streamUri: streamUri,
-      });
+      this.setStreamUri(streamUri);
       this.player.addEventListener('error', (error) => {
         if (error.detail.code === 1001) {
           this.setState({
@@ -125,7 +121,11 @@ export default class VideoSource extends React.Component {
       });
     }
   };
-
+  setStreamUri(streamUri) {
+    this.setState({
+      streamUri,
+    });
+  }
   render() {
     return this.state.src && <source src={this.state.src} />;
   }
--- a/packages/veritone-react-common/src/components/MediaPlayer/VideoSource.js
+++ b/packages/veritone-react-common/src/components/MediaPlayer/VideoSource.js
@@ -20,47 +20,31 @@ export default class VideoSource extends React.Component {
     actions: shape({
       seek: func.isRequired,
     }),
-    videoReloadTime: number.isRequired,
   };
 
   state = {
     src: null,
+    expiredSignedUri: false,
   };
 
   UNSAFE_componentWillMount() {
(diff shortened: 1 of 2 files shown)


packages/veritone-react-common/src/components/MediaPlayer/VideoSource.js
```diff
@@ -29,6 +33,51 @@ export default class VideoSource extends React.Component {
...
+      this.state.src !== prevState.src
+    ) {
+      video.load();
+    }
+    if (this.state.expiredSignedUri !== prevState.expiredSignedUri) {
+      this.setState({
+        streamUri: undefined,
+      });
+      this.loadVideo(video, src, streams, disablePreload);
+      setTimeout(() => {
```
jialu98, line 61: it seems there is api 
load(assetUri, startTime, mimeType) 
https://github.com/shaka-project/shaka-player/blob/main/lib/player.js#L1249
try it in this.loadVideo instead of using the setTimeout.

packages/veritone-react-common/src/components/MediaPlayer/VideoSource.js
```diff
@@ -29,6 +33,51 @@ export default class VideoSource extends React.Component {
...
+      !disablePreload &&
+      video &&
+      this.state.src &&
+      this.state.src !== prevState.src
+    ) {
+      video.load();
+    }
+    if (this.state.expiredSignedUri !== prevState.expiredSignedUri) {
+      this.setState({
+        streamUri: undefined,
```
jialu98, line 58: Just wondering how to get the new signed url when streamUri is expired? It seems streamUri is from streams. will streams be updated from somewhere?

packages/veritone-react-common/src/components/MediaPlayer/VideoSource.js
```diff
@@ -59,17 +114,22 @@ export default class VideoSource extends React.Component {
...
-        this.player.load(streamUri).catch(err => {
+        this.player.load(streamUri).catch((err) => {
           console.log('error loading video with shaka player', err);
         });
       }
       this.setState({
-        streamUri: streamUri
+        streamUri: streamUri,
+      });
+      this.player.addEventListener('error', () => {
```
jialu98, line 129: Is there any specific error code or message for signed url expiration? we might just want to handle specific code for signed url expiration.
https://shaka-player-demo.appspot.com/docs/api/tutorial-basic-usage.html
https://shaka-player-demo.appspot.com/docs/api/shaka.util.Error.html)

packages/veritone-react-common/src/components/MediaPlayer/VideoSource.js
```diff
@@ -29,6 +33,51 @@ export default class VideoSource extends React.Component {
...
+          actions.seek(currentTime);
+        }
+      }, 500);
+    }
+  }
+
+  componentWillUnmount() {
+    if (this.player) {
+      this.player.unload();
+    }
```
jialu98, line 72: the shaka-player adds listeners, to avoid memory leak, remove the listeners 
if (this.player) {
     await this.player.destroy();
}

packages/veritone-react-common/src/components/MediaPlayer/VideoSource.js
seta-namle: I think the code is 1001. 
image

packages/veritone-react-common/src/components/MediaPlayer/VideoSource.js
seta-namle: I think the stream URI remains the same, it will have a link like this
`"uri": "https://api.stage.us-gov-2.veritone.com/v3/stream/2360000033/dash.mpd",`
Once the media player loads the stream, it will sign new one and expired after 3 hours.

jialu98 (approved):

PR #589, opened 2024-05-03
Move JSON schemas to get.aiware.com

kevmurray:
JSON schemas used to be published to the docs.veritone.com site, but that site has been borged into support.veritone.com and can no longer host raw documents. (All pages on that site are "articles" that are dynamically generated HTML whose contents are read from a database, so it is no longer possible to host a raw JSON or text file. [Here is an example](https://support.veritone.com/s/article/000005031?language=en_US) of our "new" schema files since it has been moved to support)

The JSON schema files have been converted to be hosted at https://get.aiware.com/schemas (see https://get.aiware.com/schemas/index.html) and the release instructions updated accordingly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,14 +28,7 @@ _This is WIP and may change as we find a better process_
 9. Run `yarn publish` in each package directory as needed to push your release to NPM.
 
 ## veritone-json-schema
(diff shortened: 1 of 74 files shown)


MmtBkn (approved):
LGTM

PR #591, opened 2024-07-11
VE-3102: Update internalTaskId to type string

ndthang15:
# Description
Update `internalTaskId` to type string for safety, because although aiware taskIds are guid but maybe edge taskIds can change to not guid in the future.

Rely on the PR comment: https://github.com/veritone/aiware-core/pull/2383#discussion_r1670869478

## Related Issue

https://veritone.atlassian.net/browse/VE-3102

## How Has This Been Tested
1. Run `npm i && npm run test` in /packages/veritone-json-schemas folder.
2. Ensure that all tests are passed.


ndthang15:
Guid is string in sdk for now, so we don't need this change, and close this PR.

PR #593, opened 2024-09-09
VE-3151 release veritone-csp-generator 4.0.0

jialu98:


MmtBkn (changes requested):
This changes the sentiment operator, so let's treat it as a breaking change

MmtBkn (approved):
LGTM

PR #595, opened 2025-06-06
AION v2 spec with cognition category validation

kevmurray:
Tickets:
- VE-12831 - validation of legacy object types
- VE-12787 - licensePlate and motorVehicle support
- VE-12797 - publishing scripts


kevmurray:
Converting this back to DRAFT because there is a significant history of this package in npm that I wasn't aware of and we're not ready to support yet

iocheredko-v (approved):

PR #616, opened 2026-05-19
Ve 20021 aion v2 updates for motor vehicle track engine support

vtone-tkay:
To amend the `objectMotorVehicle` object class to accommodate fields required for Track -> Investigate integration. See [AION v2.1 Motor Vehicle Detail Changes](https://veritone.atlassian.net/wiki/spaces/VP/pages/4797890612/AION+v2.1+Motor+Vehicle+Detail+Changes)

### motorVehicle structure changes

The addition of  `confidence, `referenceId` and `tags` to the existing `motorVehicle` struct.

`confidence: type *float64` - At the object level this will represent the confidence of the detection - i.e. "How confident is the ML model that this is a vehicle?". It's useful because we may  be able to detect and track the vehicle without being confident about any of its attributes.

`referenceID: type string` - Required so that ancillary structs, such as embedding, can be linked to the `motorVehicle` object. 

`tags: type AionTag struct` - Current ML models are capable of detecting attributes that are not covered in the basic `motorVehicle` fields (e.g. vehicle damage, driver visibility). Another purpose of the tags is intended to be for alternative “guesses” as to the make and model - the current format only allows a single choice for each, but for search breadth it is better to have possible alternative labels (with their associated confidences).

### AION Schema v2.1 — motorVehicle + embedding updates

**TL;DR:** Only one schema file needs changing (`contracts.json`). The `master.json` definition of `objectResult` already has `referenceId`, `confidence`, and `tags` — the motor vehicle contract is what blocks them. Embedding requires no changes at all — the existing schema already supports `referenceId`, `vector`, and `tags`.

### Changes
`schemas/v2/contracts.json` — objectMotorVehicle contract: added referenceId, confidence, and tags to the `then.properties` allowlist (all optional). Updated `$comment` to reflect the new permitted fields.

`schemas/v2/aion/examples/oc-motorvehicle_with-reference-and-tags.json` — New valid example demonstrating the full v2.1 motor vehicle structure with referenceId, confidence, tags (using score), and vendor.

`schemas/v2/aion/examples/embedding.json` — Added a second embedding item showing the agreed pattern: referenceId + vector + tags for model/source metadata.

`package.json` — Version bumped 1.2.0 → 1.3.0.

 `packages/veritone-json-schemas/schemas/v2/index.md` - replace all mention of **AION 2.0** with **AION 2.X** to reduce user confusion. Global removal of trailing spaces on line endings. 
 
 
 ### Testing
` schemas/v2/aion/examples/oc-motorvehicle_with-reference-and-tags.json` has been added to the test suite. All 230 tests pass.


packages/veritone-json-schemas/schemas/v2/index.md
```diff
@@ -170,13 +170,13 @@ docker run --interactive --rm --volume "$PWD:/workspace" \
...
 ## Definitions
 
-Definitions referenced by other schemas. 
-These files do not contain any schema requirements themselves, 
+Definitions referenced by other schemas.
+These files do not contain any schema requirements themselves,
 only `definitions` that are referenced by other schema files.
 
-- [Master Definitions](./master.json)  
+- [Master Definitions](./master.json)
```
kevmurray, line 177: Note that whitespace at the end of markdown lines is significant. This edit changes the layout of the file. 

You seem to have deleted **all** whitespace from the ends of lines in the markdown files, which leads me to believe that you are running a tool that does this. In realtime, we also had a tool that removed whitespace from the ends of lines, but it needed to be disabled for *.md files since it breaks them.

In this case, I don't think this is a significant change

vtone-tkay: You are correct. I was mindful of, and did eyeball this (Yes it was an automated tool) but hadn't considered the `double-space = line-break` issue. 
 
The vast majority of the markdown is indented on the left so the numberof (sic) run-on word possibilities is limited by that but the concern is valid. 

I've actually set `trimTrailingWhitespace": false` for markdown now. Happy to re-instate the '  ' though - as noted - I don't see any formatting changes in the generated HTML.

kevmurray: There is a format change, but it is very minor. https://get.aiware.com/schemas/v2.0/index.html (pre change) has this:

- [Master Definitions](https://get.aiware.com/schemas/v2.0/master.json)
Contains the definitions for individual components referenced by the AION schema.

But https://get.aiware.com/schemas/pre-release/v2.1/index.html (post-change) has this

- [Master Definitions](https://get.aiware.com/schemas/pre-release/v2.1/master.json) Contains the definitions for individual components referenced by the AION schema.

This is a minor enough change that I don't think it's worth going back to "fix". My comment was mostly about warning you about line-endings in Markdown... it doesn't matter here so much but I've had situations where it was much more significant and made the document harder to read instead of being a minor format change.

kevmurray (approved):
Sorry, I thought I approved this. It may not count because it's still in Draft?

UPDATE: Nope, draft doesn't matter... I must have simply forgotten to click "Approve"

